PR #39, opened 2024-12-04
feat: request grant confirmation for config updates

lengyel-arpad85:
## Changes proposed

- add open-payments library
- request grant confirmation when creating, deleting or updating a config
- after first grant confirmation, we save the wallet address for the duration of the session, to prevent prompting user at every action


frontend/app/lib/session.ts
```diff
@@ -0,0 +1,16 @@
+import { createCookieSessionStorage } from '@remix-run/node'
+
+const { getSession, commitSession, destroySession } =
+  createCookieSessionStorage({
+    cookie: {
+      name: 'wmtools-session',
+      httpOnly: true,
```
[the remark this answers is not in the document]
sidvishnoi: And if there's enough such files, a `.server` directory?
Don't want to accidentally expose secrets to client in case of a bad import.

frontend/app/routes/create.$type.tsx
```diff
@@ -298,87 +415,113 @@ export async function action({ request, params }: ActionFunctionArgs) {
...
+      try {
+        const walletAddress = await getValidWalletAddress(ownerWalletAddress)
+        session.set('wallet-address', walletAddress)
+
+        const quote = await fetchQuote({
+          senderAddress: walletAddress,
+          receiverAddress: walletAddress,
+          amount: 1, // 0 value fails and fee is calculated from this
+          note: 'Publisher Tools wallet verification'
+        })
```
sidvishnoi, line 463: Do we need a quote? Can we not create a grant with an implicit quote (pass `debitAmount` instead of `quote`)? One less request from our server.

frontend/app/lib/open-payments.server.ts
```diff
@@ -0,0 +1,318 @@
...
+  type AuthenticatedClient,
+  type PendingGrant,
+  type Quote,
+  type WalletAddress,
+  createAuthenticatedClient,
+  isFinalizedGrant,
+  isPendingGrant
+} from '@interledger/open-payments'
+import { createId } from '@paralleldrive/cuid2'
+import { randomUUID } from 'crypto'
```
sidvishnoi, line 11: Nit: Maybe use the global `crypto.randomUUID()`? That way we don't have to over-rely on Node env in all the files. Helps with migration.

frontend/app/routes/create.$type.tsx
```diff
@@ -298,87 +415,113 @@ export async function action({ request, params }: ActionFunctionArgs) {
...
+
+    if (!validForWallet || validForWallet !== ownerWalletAddress) {
+      try {
+        const walletAddress = await getValidWalletAddress(ownerWalletAddress)
+        session.set('wallet-address', walletAddress)
+
+        const quote = await fetchQuote({
+          senderAddress: walletAddress,
+          receiverAddress: walletAddress,
+          amount: 1, // 0 value fails and fee is calculated from this
```
sidvishnoi, line 461: This amount `1` is valid only for test wallet I think? Different wallets can take a different fee.

frontend/app/lib/session.ts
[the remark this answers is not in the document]
lengyel-arpad85: yes server component, 
renamed the file

frontend/app/routes/create.$type.tsx
lengyel-arpad85: fetch quote was removed

frontend/app/lib/server/open-payments.server.ts
```diff
@@ -0,0 +1,319 @@
...
+import { createId } from '@paralleldrive/cuid2'
+import { toWalletAddressUrl } from '../utils.js'
+
+async function createClient() {
+  return await createAuthenticatedClient({
+    keyId: process.env.OP_KEY_ID!,
+    privateKey: Buffer.from(process.env.OP_PRIVATE_KEY!, 'base64'),
+    walletAddressUrl: process.env.OP_WALLET_ADDRESS!
+  })
+}
```
sidvishnoi, line 19: Can cache this client.

```suggestion
let client
async function createClient() {
  client ??= await createAuthenticatedClient({
    keyId: process.env.OP_KEY_ID!,
    privateKey: Buffer.from(process.env.OP_PRIVATE_KEY!, 'base64'),
    walletAddressUrl: process.env.OP_WALLET_ADDRESS!
  })
  return client
}
```

frontend/app/routes/create.$type.tsx
```diff
@@ -75,35 +101,53 @@ export default function Create() {
...
+    version: string
+  ) => {
+    const formData = new FormData()
+    formData.append('intent', intent)
+    formData.append('version', version)
+    formData.append('fullconfig', JSON.stringify(configArray))
+    const defaultSet = configArray.default as unknown as Record<string, string>
+    Object.keys(defaultSet).map((key) => {
+      formData.append(key, defaultSet[key])
+    })
```
sidvishnoi, line 145: Recommend setting required-ish fields later, in case some key from `defaultSet` overrides intent/version/fullconfig in future. And using `formData.set` instead of append so there's a single key (otherwise might need to use `getAll` in backend).

```suggestion
    const defaultSet = configArray.default as unknown as Record<string, string>
    Object.keys(defaultSet).map((key) => {
      formData.set(key, defaultSet[key])
    })
    formData.set('intent', intent)
    formData.set('version', version)
    formData.set('fullconfig', JSON.stringify(configArray))
```

frontend/app/routes/create.$type.tsx
```diff
@@ -115,25 +159,82 @@ export default function Create() {
...
+    switch (type) {
+      case 'confirm':
+        title = `Are you sure you want to remove ${selectedVersion}?`
+        onConfirm = onConfirmRemove
+        break
+      case 'wallet-ownership':
+        title = `Please confirm you are owner of 
+                <span class="flex w-full justify-center text-center">
+                  ${walletAddress.id}
+                </span>`
```
sidvishnoi, line 196: Can we pass JSX nodes instead of `dangerouslySetInnerHTML`? `title` would then be `ReactNode` or `string | ReactElement`.

frontend/app/lib/server/open-payments.server.ts
```diff
@@ -0,0 +1,319 @@
...
+      {
+        url: walletAddress.authServer
+      },
+      {
+        access_token: {
+          access: [
+            {
+              identifier: walletAddress.id,
+              type: 'outgoing-payment',
+              actions: ['create', 'read', 'list'],
```
sidvishnoi, line 196: This might be enough? Mostly, `'list'` be unnecessary.
```suggestion
              actions: ['create'],
```

frontend/app/lib/server/open-payments.server.ts
lengyel-arpad85: this fails typecheck, so I reverted it
the original function is from other implementations, it can live as is for now

frontend/app/routes/create.$type.tsx
lengyel-arpad85: there's no overlap regarding the data, but easy enough to change the order, so it's done

sidvishnoi (approved):
Haven't tested locally, but code looks fine.

backend/src/services/session.ts
```diff
@@ -0,0 +1,17 @@
+import { createCookieSessionStorage } from '@remix-run/node'
+
+export const SESSION_COOKIE_SECRET_KEY =
+  process.env.SESSION_COOKIE_SECRET_KEY || 'supersecretilpaystring'
+
+const { getSession, commitSession, destroySession } =
+  createCookieSessionStorage({
+    cookie: {
+      name: 'wmtools-session',
```
DarianM, line 9: do you think we should add a maxAge or delete the wmtools-session cookie when the session/tab closes?
basically asking the user for a new grant confirmation over a period of time or after the tab closes.

only session gets cleared after a session but not this

lengyel-arpad85: sadly there is no way of detecting when tab closes (browser event `onunload` fires even when navigating to a different domain, so it would remove the cookie everytime we go to IDP to confirm the grant)
the only solution I see is add maxAge, just need to decide how long that should be ? 10, 30  minutes ? 1h ?

PR #52, opened 2025-03-07
feat: button embedded component

### LINKED ISSUE #4: Create button as embeddable component
as a next step after v1, we should have a button as an option for users to customize / create.
Clicking the button by a visitor on a website should make an overlay over window / screen and display the same payment option as the widget

DarianM:
## Change proposed

1. Button overlay:
   - embedded button opens overlay with iframe interledgerpay
   - button styling inheritance

2. Tooltip Implementation:
   - configurable show/hide behavior
   - consistent styling with button theme

3. Overlay improvements:
   - added proper backdrop handling
   - structured content with header/body/footer
   - close button functionality


frontend/script/index.tsx
```diff
@@ -123,16 +123,15 @@ const drawElement = (
   walletAddress: string,
   config: Config
 ) => {
-  const { shadowHost, shadowRoot } = createShadowDOM()
-
   // add payment pointer / wallet address to target website first
-  // so we have something to check against when displaying the banner
+  // pay into this address when displaying embedded element
```
lengyel-arpad85, line 127: no, the payment pointer is added to the host page, but has nothing to do w payments on the tools side
we are using it to determine if user has WM extension ... and  as a secondary reason this way website has at least 1 PP that the extansion can use to pay into
`// check if user / visitor has monetization

  const monetizationLinks = document.querySelector<HTMLLinkElement>(
    'link[rel=monetization]'
  )
  if (
    (monetizationLinks && monetizationLinks.relList.supports('monetization')) ||
    closedByUser
  ) {
    // prevent element being created, if the extension is installed
    return
  }`

frontend/public/css/overlay.css
```diff
@@ -0,0 +1,108 @@
...
+/* widget Powered By (used in button footer) */
+._wm_tools_widget_poweredby {
+  display: flex;
+  margin-top: 1.5rem;
+  cursor: pointer;
+}
+
+._wm_tools_widget_poweredby img {
+  height: 2.5rem;
+}
```
lengyel-arpad85, line 108: please make sure you are not altering the existing elements, this was discussed and agreed upon that widget footer has the powered by design with no margins / padding "glued" to the sides and bottom. it needs to look like this:

Screenshot from 2025-03-10 10-19-03

frontend/script/index.tsx
```diff
@@ -261,23 +266,213 @@ const drawBanner = (config: Config) => {
...
+  overlayBody.className = '_wm_tools_button_overlay_body'
+
+  const iframe = document.createElement('iframe')
+  iframe.id = 'button_ilpay_iframe'
+  iframe.className = '_wm_tools_button_overlay_iframe'
+  iframe.src = generateIframeUrl()
+  iframe.title = 'Payment Preview'
+  iframe.setAttribute('loading', 'eager')
+
+  overlayBody.appendChild(iframe)
```
lengyel-arpad85, line 337: guessing this is missing the css parameter

Screenshot from 2025-03-10 10-26-33

frontend/script/index.tsx
```diff
@@ -261,23 +266,213 @@ const drawBanner = (config: Config) => {
...
+
+  overlay.appendChild(backdrop)
+  overlay.appendChild(overlayContent)
+
+  return overlay
+}
+
+const setupButtonPlaceholders = (
+  config: Config,
+  buttonShadowRoot: ShadowRoot
```
lengyel-arpad85, line 360: the buttons need to be placed into the shadowDOM in order for the host page doesn't interfere w them, if you check source in the browser, the shadow-root is missing please check / fix

Screenshot from 2025-03-10 10-30-02
Screenshot from 2025-03-10 10-30-44

lengyel-arpad85 (changes requested):

lengyel-arpad85 (commented):
also, the tooltip on the embed side needs fixing

Screenshot from 2025-03-10 10-34-45

frontend/app/components/InfoWithTooltip.tsx
```diff
@@ -5,9 +5,7 @@ import {
   useEffect,
   useState
 } from 'react'
-import Tippy from '@tippyjs/react'
 import { Info } from './icons.js'
-import 'tippy.js/dist/tippy.css'
```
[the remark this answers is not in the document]
DarianM: we can add tippy back since now i see it's used in multiple places, but still we need custom html when adding the button into the shadow dom on the host page

PR #63, opened 2025-03-13
fix: data sanitization

DarianM:
### proposed changes

- replace innerHTML with createElement; use document.createTextNode for text context and set attributes directly
- refine wallet address validation in corresponding inputs; 
- add html sanitization on backend for create and save config

### DIFF
--- a/backend/src/controllers/tools.ts
+++ b/backend/src/controllers/tools.ts
@@ -37,7 +37,7 @@ export const createUserConfig = async (req: Request, res: Response) => {
       throw 'Wallet address is required'
     }
     const defaultData = await getDefaultData()
-    const defaultDataContent: ConfigVersions[keyof ConfigVersions] =
+    const defaultDataContent: ConfigVersions['default'] =
       JSON.parse(defaultData).default
     defaultDataContent.walletAddress = decodeURIComponent(
       `https://${data.walletAddress}`
@@ -222,7 +222,7 @@ const sanitizeConfigFields = <T extends Partial<SanitizedFields>>(
 
   for (const field of textFields) {
     const value = config[field]
-    if (value) {
+    if (typeof value === 'string' && value) {
       const decoded = he.decode(value)
       const sanitizedText = sanitizeHtml(value, {
         allowedTags: [],
@@ -232,7 +232,7 @@ const sanitizeConfigFields = <T extends Partial<SanitizedFields>>(
         }
       })
       if (sanitizedText !== decoded) {
-        throw new Error(`Invalid HTML in field: ${field}`)
+        throw new Error(`HTML not allowed in field: ${field}`)
       }
 
       config[field] = sanitizedText
@@ -240,7 +240,7 @@ const sanitizeConfigFields = <T extends Partial<SanitizedFields>>(
   }
 
   for (const field of htmlFields) {
-    if (config[field]) {
+    if (typeof config[field] === 'string' && config[field]) {
       const decoded = he.decode(config[field].replace(/&nbsp;/g, '').trim())
       const sanitizedHTML = sanitizeHtml(decoded, {
         allowedTags: [],
--- a/backend/src/controllers/tools.ts
+++ b/backend/src/controllers/tools.ts
@@ -37,9 +37,8 @@ export const createUserConfig = async (req: Request, res: Response) => {
       throw 'Wallet address is required'
     }
     const defaultData = await getDefaultData()
-    const defaultDataContent: ConfigVersions[keyof ConfigVersions] & {
-      walletAddress: string
-    } = JSON.parse(defaultData).default
+    const defaultDataContent: ConfigVersions[keyof ConfigVersions] =
+      JSON.parse(defaultData).default
     defaultDataContent.walletAddress = decodeURIComponent(
       `https://${data.walletAddress}`
     )
@@ -205,11 +204,9 @@ export const getUserConfigByTag = async (req: Request, res: Response) => {
 const sanitizeConfigFields = <T extends Partial<SanitizedFields>>(
   config: T
 ): T => {
-  const fieldsToSanitize: Array<keyof SanitizedFields> = [
+  const textFields: Array<keyof SanitizedFields> = [
     'bannerTitleText',
-    'bannerDescriptionText',
     'widgetTitleText',
-    'widgetDescriptionText',
     'widgetButtonText',
     'buttonText',
     'buttonDescriptionText',
@@ -218,14 +215,39 @@ const sanitizeConfigFields = <T extends Partial<SanitizedFields>>(
     'version'
   ]
 
-  for (const field of fieldsToSanitize) {
+  const htmlFields: Array<keyof SanitizedFields> = [
+    'bannerDescriptionText',
+    'widgetDescriptionText'
+  ]
+
+  for (const field of textFields) {
     const value = config[field]
     if (value) {
       const decoded = he.decode(value)
-      const sanitized = sanitizeHtml(decoded, {
-        allowedAttributes: {}
+      const sanitizedText = sanitizeHtml(value, {
+        allowedTags: [],
+        allowedAttributes: {},
+        textFilter(text) {
+          return he.decode(text)
+        }
       })
-      const decodedSanitized = he.decode(sanitized)
+      if (sanitizedText !== decoded) {
+        throw new Error(`Invalid HTML in field: ${field}`)
+      }
+
+      config[field] = sanitizedText
+    }
+  }
+
+  for (const field of htmlFields) {
+    if (config[field]) {
+      const decoded = he.decode(config[field].replace(/&nbsp;/g, '').trim())
+      const sanitizedHTML = sanitizeHtml(decoded, {
+        allowedTags: [],
+        allowedAttributes: {},
+        allowProtocolRelative: false
+      })
+      const decodedSanitized = he.decode(sanitizedHTML)
       // compare decoded versions to check for malicious content
       if (decodedSanitized !== decoded) {
         throw new Error(`Invalid HTML in field: ${field}`)
--- a/backend/src/controllers/tools.ts
+++ b/backend/src/controllers/tools.ts
@@ -1,3 +1,4 @@
+import sanitizeHtml from 'sanitize-html'
 import type { Request, Response } from 'express'
 import { PutObjectCommand, GetObjectCommand } from '@aws-sdk/client-s3'
 import _ from 'underscore'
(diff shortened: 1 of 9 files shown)


lengyel-arpad85 (changes requested):

backend/src/controllers/tools.ts
```diff
@@ -184,6 +206,34 @@ export const getUserConfigByTag = async (req: Request, res: Response) => {
...
+  ]
+
+  for (const field of fieldsToSanitize) {
+    if (config[field]) {
+      const sanitized = sanitizeHtml(config[field], {
+        allowedTags: ['b', 'i', 'em', 'strong']
+      })
+      if (sanitized !== config[field]) {
+        throw new Error(`Invalid HTML in field: ${field}`)
+      }
```
sidvishnoi, line 230: Careful with encoding of special chars, such as `&`.

frontend/app/lib/validate.server.ts
```diff
@@ -4,9 +4,59 @@ import { CornerType, PositionType, SlideAnimationType } from './types.js'
...
+    if (response.status === 404) {
+      throw new Error('This wallet address does not exist.')
+    }
+    throw new Error('Failed to fetch wallet address.')
+  }
+
+  const msgInvalidWalletAddress = 'Provided URL is not a valid wallet address.'
+  await response.json().catch((error) => {
+    throw new Error(msgInvalidWalletAddress, { cause: error })
+  })
```
sidvishnoi, line 55: Should also validate whether JSON corresponds to a wallet address. Code from extension:

https://github.com/interledger/web-monetization-extension/blob/6c24ed9725f25426749e32956e6d3a23e7f84d50/src/shared/helpers.ts#L37-L50

frontend/app/lib/validate.server.ts
```diff
@@ -4,9 +4,59 @@ import { CornerType, PositionType, SlideAnimationType } from './types.js'
...
+    })
 })
 
+function toWalletAddressUrl(s: string): string {
+  return s.startsWith('$') ? s.replace('$', 'https://') : s
+}
+
+const isValidWalletAddress = async (
+  walletAddressUrl: string
+): Promise<void> => {
```
sidvishnoi, line 39: I'd expect a `is*` function to return a boolean.

```suggestion
): Promise<boolean> => {
```

frontend/script/index.tsx
```diff
@@ -232,32 +208,77 @@ const drawBanner = (config: Config) => {
...
+    return 'https://addons.mozilla.org/en-US/firefox/addon/web-monetization-extension/'
+  } else if (
+    userAgent.includes('Chrome') &&
+    !userAgent.includes('Edg') &&
+    !userAgent.includes('OPR')
+  ) {
+    return 'https://chromewebstore.google.com/detail/web-monetization-extensio/oiabcfomehhigdepbbclppomkhlknpii'
+  } else if (userAgent.includes('Edg')) {
+    return 'https://microsoftedge.microsoft.com/addons/detail/web-monetization-extensio/imjgemgmeoioefpmfefmffbboogighjl'
+  }
```
sidvishnoi, line 266: I think this should happen before Chrome? Like if `Edg` we're sure it's edge?

sidvishnoi: Might we worth while to check https://www.npmjs.com/package/ua-parser-js as we can return extension URL from server as well if needed?

backend/src/controllers/tools.ts
DarianM: fixed special chars. we'll no longer see `&amp;` on pages

lengyel-arpad85 (approved):

backend/src/controllers/tools.ts
```diff
@@ -184,6 +202,41 @@ export const getUserConfigByTag = async (req: Request, res: Response) => {
...
+    'version'
+  ]
+
+  for (const field of fieldsToSanitize) {
+    const value = config[field]
+    if (value) {
+      const decoded = he.decode(value)
+      const sanitized = sanitizeHtml(decoded, {
+        allowedAttributes: {}
+      })
```
sidvishnoi, line 227: In one of my earlier projects, I've used it like this. Let me know if that gives better results.

```ts
// for plain text fields, such as title where we don't want any HTML
const sanitizedText = sanitizeHtml(input, {
  allowedTags: [],
  allowedAttributes: {},
  textFilter(text) {
    return decode(text); // unescape
  },
});

// for fields where we want some restricted HTML
const input = rawInput.replace(/&nbsp;/g, '').trim();
const sanitizedHTML = sanitizeHtml(input, {
  allowedTags: ['b', 'i', 'em', 'strong', 'a'],
  allowedAttributes: {
    a: ['href'],
  },
  allowedSchemes: ['http', 'https'],
  allowProtocolRelative: false,
});
```

backend/src/controllers/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  walletAddress: string
+  fullconfig: string // JSON stringified object containing all versions
+  version: string
+  // ... other fields
+}
+
+export interface ConfigVersions {
+  [key: string]: {
+    // button config
+    buttonFontName: string
```
sidvishnoi, line 17: Nit: I'd create a separate type `ConfigVersion` to allow reuse and then do:

```ts
interface ConfigVersions {
  [key: string]: ConfigVersion
}
```

backend/src/controllers/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  bannerDescriptionText?: string
+  widgetTitleText?: string
+  widgetDescriptionText?: string
+  widgetButtonText?: string
+  buttonText?: string
+  buttonDescriptionText?: string
+  walletAddress?: string
+  tag?: string
+  version?: string
+}
```
sidvishnoi, line 69: I think they'll all be strings only? If so, maybe something like:

```ts
type Fields = 'bannerTitleText' | 'bannerDescriptionText' // | ...
type SanitizedFields = Partial<Record<Fields, string>>
```

frontend/app/lib/validate.server.ts
```diff
@@ -1,12 +1,66 @@
...
+        const parsed = new URL(toWalletAddressUrl(url))
+        if (parsed.protocol !== 'https:') {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: 'Wallet address must use HTTPS protocol'
+          })
+          return
+        }
+
+        await isValidWalletAddress(parsed.toString())
```
sidvishnoi, line 24: I think errors from this one should handled differently. Like we have more detailed errors available (not found etc.), and this isn't always "invalid wallet address format" error.

frontend/script/index.tsx
```diff
@@ -232,32 +208,77 @@ const drawBanner = (config: Config) => {
...
-  element.insertAdjacentHTML(
-    'beforeend',
-    `<span class="_wm_link">${getWebMonetizationLink()}</span>`
-  )
+  bannerHeader.appendChild(closeButton)
+  element.appendChild(bannerHeader)
+
+  // description text
+  const descriptionSpan = document.createElement('span')
+  const descriptionText = document.createTextNode(config.bannerDescriptionText)
```
sidvishnoi, line 234: We do allow `<em>` like tags in `bannerDescriptionText`, right? Or not yet? Text node will swallow those elements?

frontend/script/index.tsx
```diff
@@ -232,32 +208,77 @@ const drawBanner = (config: Config) => {
...
+const getWebMonetizationLinkHref = () => {
+  const userAgent = navigator.userAgent
+  if (userAgent.includes('Firefox')) {
+    return 'https://addons.mozilla.org/en-US/firefox/addon/web-monetization-extension/'
+  } else if (
+    userAgent.includes('Chrome') &&
+    !userAgent.includes('Edg') &&
+    !userAgent.includes('OPR')
+  ) {
+    return 'https://chromewebstore.google.com/detail/web-monetization-extensio/oiabcfomehhigdepbbclppomkhlknpii'
```
sidvishnoi, line 263:
```suggestion
    return 'https://chromewebstore.google.com/detail/web-monetization/oiabcfomehhigdepbbclppomkhlknpii'
```

frontend/script/index.tsx
```diff
@@ -232,32 +208,77 @@ const drawBanner = (config: Config) => {
...
+  } else if (userAgent.includes('Edg')) {
+    return 'https://microsoftedge.microsoft.com/addons/detail/web-monetization-extensio/imjgemgmeoioefpmfefmffbboogighjl'
```
sidvishnoi, line 265:
```suggestion
    return 'https://microsoftedge.microsoft.com/addons/detail/web-monetization/imjgemgmeoioefpmfefmffbboogighjl'
```

frontend/script/index.tsx
```diff
@@ -232,32 +208,77 @@ const drawBanner = (config: Config) => {
...
+
+const getWebMonetizationLinkText = () => {
+  const userAgent = navigator.userAgent
+  if (
+    userAgent.includes('Firefox') ||
+    userAgent.includes('Chrome') ||
+    userAgent.includes('Edg')
+  ) {
+    return 'Get the extension'
+  }
```
sidvishnoi, line 278: Can use result of `getWebMonetizationLinkHref` here instead (i.e. essentially using same userAgent handling).

backend/src/controllers/types.ts
sidvishnoi: Alternatively, `Pick` from `ConfigVersion`?

frontend/script/index.tsx
DarianM: yes, creeateTextNode will output everything as plain text, including <em> tags, even if we allow such such as `<em>` `<b>` .  I remember your suggestion  editorjs to handle such elements in description text

backend/src/controllers/tools.ts
DarianM: we can allow tags, even if we don't properly handle them.
but i'm really reserved on allowing `a` tags, even with only `href `attr. wdyt ?

sidvishnoi: Let's user add custom links in description? We can for first pass allow only plain text I think.

DarianM: yes, we can use this for banner description text even now, in preparation of displaying allowed tags properly

frontend/app/lib/validate.server.ts
```diff
@@ -1,10 +1,34 @@
...
 const rangeError = { message: 'Value has to be between 16 and 24' }
 
 export const walletSchema = z.object({
-  walletAddress: z.string().min(1, { message: 'Wallet address is required' })
+  walletAddress: z
+    .string()
+    .min(1, { message: 'Wallet address is required' })
+    .superRefine(async (url, ctx) => {
+      if (url.length === 0) return
+
```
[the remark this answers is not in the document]
DarianM: The problem I have with zod is that .min(1) and .superRefine are parallel validations, not sequential. Both validations run independently, and it collects all validation errors. I do not want to show 2 message errors if I pass an empty string to schema

PR #68, opened 2025-03-19
feat: add presets for revshare and linktag

DarianM:
### Requested changes

These changes belong to the initiative to move rev-share & linktag generators under publisher tools.
Add `description` and `tooltip` text in tools home page for both linktag and rev-share


lengyel-arpad85 (approved):

frontend/app/lib/tooltips.ts
```diff
@@ -7,5 +7,9 @@ export const tooltips = {
     'This widget enables one-time payments for your website. It appears in the lower right corner, visible to all visitors regardless of their Web Monetization status. The title, text, and styling of the displayed message are customizable.',
   scriptModal:
     'Copy your script and insert it just before the closing </body> tag on your website, or place it within a script type element if you\'re using a site management software (ex: WordPress, etc). <br />Be sure to review and select all the options in the "Include" section, that you want to display on your website.',
+  linkTagGenerator:
+    'Generate and customize your Web Monetization <link> tag to monetize your web pages. The generator helps you create the correct tag with your payment pointer and validates the syntax, ensuring proper implementation of Web Monetization in your HTML documents.',
+  revShareGenerator:
+    'Create a customized revenue sharing configuration that distributes Web Monetization earnings between multiple payment pointers. The generator helps you set weighted percentages for each recipient and validates the payment pointer format, ensuring fair distribution of streaming payments.',
```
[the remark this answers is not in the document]
DarianM: no it's not, still looks good with some line break elements in between. thanks!

frontend/app/lib/presets.ts
```diff
@@ -135,6 +135,26 @@ export const availableTools = [
     description:
       'You can place a custom button with a short tooltip on your site that triggers a payment option in a full-page overlay, offering a convenient way for visitors to support your work or content.'
   },
+  {
+    enabled: true,
+    title: 'Link',
+    image: 'link_representation.svg',
+    bgColor: 'from-wm-red to-wm-red-fade',
+    link: `https://webmonetization.org/tools/link-tag/`,
```
sidvishnoi, line 143: Do this open in a new tab? If not, they'll open within iframe and might end up showing 2x headers etc. on page.

DarianM: no new tab. because tools ends up in an iframe, clicking the `link` would redirect in iframe instead of main parent window. very good point.

frontend/app/components/ButtonOrLink.tsx
```diff
@@ -49,7 +49,9 @@ export const ButtonOrLink = forwardRef<
   const isLink = typeof to !== 'undefined' || typeof href !== 'undefined'
 
   if (isLink) {
-    return <AnchorOrLink href={href} to={to} ref={ref} {...props} />
+    return (
+      <AnchorOrLink href={href} to={to} ref={ref} {...props} target="_parent" />
+    )
```
sidvishnoi, line 54: Would be better to do this in `TypeCard` instead, passing this `target="_parent"` to `Button` there. `ButtonOrLink` is too generic to add this behavior - it can bite us in future.

PR #82, opened 2025-03-24
fix: delete config route

DarianM:
### Proposed changes

- created a dedicated DELETE endpoint
- handles config version deletion server-side
- Returns updated config after deletion
- Maintains grant/wallet validation

### DIFF
--- a/backend/src/controllers/tools.ts
+++ b/backend/src/controllers/tools.ts
@@ -222,6 +222,50 @@ export const getUserConfigByTag = async (req: Request, res: Response) => {
   }
 }
 
+export const deleteUserConfigVersion = async (req: Request, res: Response) => {
+  try {
+    const { walletAddress, version } = req.params
+    const cookieHeader = req.headers.cookie
+    const session = await getSession(cookieHeader)
+
+    const validForWallet = session?.get('validForWallet')
+    if (!walletAddress) {
+      throw 'Wallet address is required'
+    }
+    if (!session || validForWallet !== walletAddress) {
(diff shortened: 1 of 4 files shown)


backend/src/controllers/tools.ts
```diff
@@ -222,6 +222,50 @@ export const getUserConfigByTag = async (req: Request, res: Response) => {
...
+      await s3.send(new PutObjectCommand(extendedParams))
+    }
+
+    res.status(200).send(existingConfig)
+  } catch (err) {
+    console.log(err)
+    const message =
+      err === 'Cannot delete default version'
+        ? err
+        : 'An error occurred when deleting config version'
```
lengyel-arpad85, line 264: we have errors defined here:
backend/src/services/errors.ts
maybe put this in same place, 
not a blocker

backend/src/routes/tools.ts
```diff
@@ -13,6 +14,7 @@ const userRoutes = (router: Router) => {
   router.get('/tools/:id/:tag', getUserConfigByTag)
   router.post('/tools', createUserConfig)
   router.put('/tools', saveUserConfig)
+  router.delete('/config/:walletAddress/:version', deleteUserConfigVersion)
```
lengyel-arpad85, line 17: routes all use `/tools` make it consistent please, either rename all to config or this should be tools also

backend/src/controllers/tools.ts
DarianM: thanks for pointing out! will move these errors in a different PR 👍

PR #103, opened 2025-05-12
fix(frontend): make it work with cloudflare-pages adapter

DarianM:
## Proposed changes

* added a cloudflare pages function handler in `frontend/functions/[[path]].ts` to handle requests using the remix server build. 
* updated `package.json` to include `@remix-run/cloudflare-pages` as a dependency and added a `deploy` script for deploying to cloudflare pages.
   * Updated related dependencies too.

### cleanup:
* removed `itty-router` from `package.json` as it is no longer being used.


frontend/worker-configuration.d.ts
sidvishnoi, line 1: Add this to `.gitignore` as well. I think can do in top-level .gitignore

frontend/package.json
```diff
@@ -6,13 +6,15 @@
...
     "dev": "remix vite:dev --port 5101",
-    "preview": "wrangler pages dev ./build/client --port=5100 --binding SCRIPT_FRONTEND_URL=http://localhost:5100/ --compatibility-date=2024-09-23 --compatibility-flag=nodejs_compat"
+    "preview": "pnpm run build && wrangler pages dev ./build/client",
+    "deploy": "pnpm run build && wrangler pages deploy"
   },
   "dependencies": {
     "@headlessui/react": "^1.7.19",
     "@interledger/open-payments": "^6.13.2",
     "@paralleldrive/cuid2": "^2.2.2",
     "@remix-run/cloudflare": "^2.16.5",
```
sidvishnoi, line 16: Let's update this package as well, keeping in line with `@remix-run/cloudflare-pages`

sidvishnoi: Along with other remix packages?

frontend/package.json
```diff
@@ -6,13 +6,15 @@
   "scripts": {
     "build": "remix vite:build",
     "dev": "remix vite:dev --port 5101",
-    "preview": "wrangler pages dev ./build/client --port=5100 --binding SCRIPT_FRONTEND_URL=http://localhost:5100/ --compatibility-date=2024-09-23 --compatibility-flag=nodejs_compat"
+    "preview": "pnpm run build && wrangler pages dev ./build/client",
```
sidvishnoi, line 9:
```suggestion
    "preview": "pnpm run build && wrangler pages dev",
```

And add to `wrangler.toml`:
```toml
pages_build_output_dir="./build/client"
```

frontend/wrangler.toml
```diff
@@ -1,13 +1,6 @@
...
+compatibility_flags = ["nodejs_compat"]
 compatibility_date = "2024-09-23"
 
 [dev]
-port = 5100
-
-[site]
-bucket = "./public"
-
-[vars]
```
sidvishnoi, line 12: I think we want to define the non-secret vars here. `.dev.vars` will let us override during dev, but otherwise, we'll use production ones defined here.

frontend/wrangler.toml
```diff
@@ -4,3 +4,9 @@ compatibility_date = "2024-09-23"
 
 [dev]
 port = 5101
+
+[vars]
+SCRIPT_API_URL = "https://localhost:5101/"
+SCRIPT_FRONTEND_URL = "http://localhost:5101/"
```
sidvishnoi, line 11: Why does the frontend need `SCRIPT_FRONTEND_URL`?

DarianM: as redirect `uri` after interactive grant, that's the authentication part.
but certainly we can remove `SCRIPT_API_URL` or `SCRIPT_FRONTEND_URL`,

DarianM: or can it be we do not need a env variables for this redirect and can get rid of both

frontend/app/lib/utils.ts
```diff
@@ -265,12 +265,12 @@ export const encodeAndCompressParameters = async (params: string) => {
...
 
-  // eslint-disable-next-line no-constant-condition
-  while (true) {
+  let reading = true
+  while (reading) {
     const { done, value } = await reader.read()
     if (done) {
       reader.releaseLock()
-      break
+      reading = false
```
sidvishnoi, line 273: I would've preferred the `break` one (even if it needs a eslint ignore).

Alternatively, as a follow up, see if it's possible to use `for await (const value of reader.something())`.

frontend/functions/[[path]].ts
Comment (sidvishnoi) [inline frontend/functions/[[path]].ts:1]: For the record: This was the main file that we needed for this PR.

sidvishnoi (approved):

frontend/wrangler.toml
```diff
@@ -1,13 +1,11 @@
...
+compatibility_flags = ["nodejs_compat"]
 compatibility_date = "2024-09-23"
+pages_build_output_dir = "./build/client"
 
 [dev]
-port = 5100
-
-[site]
-bucket = "./public"
+port = 5102
```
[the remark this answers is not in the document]
DarianM: sure! i have no preferences here. comfortable with 3000

frontend/wrangler.toml
```diff
@@ -4,7 +4,7 @@ compatibility_date = "2024-09-23"
 pages_build_output_dir = "./build/client"
 
 [dev]
-port = 5102
+port = 3001
```
sidvishnoi, line 7: Did you mean:
```suggestion
port = 3000
```

Or it should be different from `vite:dev` one?

DarianM: now that we removed the need of `SCRIPT_FRONTEND_URL`
we can have diff ports for `dev` and `preview`
just to not create confusion between scripts.
or have both scripts running in parallel

frontend/app/lib/utils.ts
DarianM: something to do as a follow up for sure

PR #108, opened 2025-05-16
fix: polyfill SDK crypto module as unsupported in Cloudflare

DarianM:
### Description
OpenPayments uses nodejs environment for keys cryptography, but cloudflare supports Web API platform for it's workers, (i.e. Web Crypto API) creating SDK incompatibility when deploying.
See https://github.com/interledger/open-payments/issues/602

## Open Payments Integration Enhancements
* support for cryptographic signing of HTTP requests and creating headers to bypass  OP's `crypto` dependency when creating the authenticated client.

## Build Configuration Updates
* Added a `resolve.alias` entry to use `crypto-browserify` as a replacement for the nodejs `crypto` module in the browser and updated the build target to `esnext` with conditional source maps for development.


embed/vite.config.ts
```diff
@@ -5,7 +5,17 @@ export default defineConfig(({ mode }) => {
...
+      'import.meta.env.VITE_SCRIPT_FRONTEND_URL': JSON.stringify(
+        'https://webmonetization.org.pages.dev/tools/'
+      ),
+      'import.meta.env.VITE_SCRIPT_API_URL': JSON.stringify(
+        'https://api.webmonetization.org.pages.dev/'
+      ),
+      'import.meta.env.VITE_SCRIPT_ILPAY_URL': JSON.stringify(
+        'https://interledgerpay.com/extension/'
+      )
+    },
```
sidvishnoi, line 18: Should keep these in env or some constants file. Specially `VITE_SCRIPT_API_URL` as it's different during dev.

sidvishnoi: Still, let's move this out of this PR.

frontend/app/utils/open-payments.server.ts
```diff
@@ -1,20 +1,81 @@
...
+
+      const initialRequest = request.clone()
+
+      const headers = await createHeaders({
+        request: {
+          method: request.method,
+          url: request.url,
+          headers: JSON.parse(
+            JSON.stringify(Object.fromEntries(request.headers))
+          ),
```
sidvishnoi, line 56:
```suggestion
          headers: new Headers(request.headers),
```

Will this work?

frontend/app/utils/open-payments.server.ts
```diff
@@ -108,21 +169,7 @@ async function createOutgoingPaymentGrant(
...
-      console.error({
-        errorCode,
-        errorStatus,
-        errorMessage,
-        url: redirectUrl,
-        walletAddress
-      })
-
+    .catch(() => {
       throw new Error('Could not retrieve outgoing payment grant.')
```
sidvishnoi, line 173:
```suggestion
    .catch((err) => {
      throw new Error('Could not retrieve outgoing payment grant.', { cause: err })
```

frontend/app/utils/open-payments.server.ts
```diff
@@ -1,20 +1,81 @@
...
+}
+
+interface ContentHeaders {
+  'Content-Digest': string
+  'Content-Length': string
+  'Content-Type': string
+}
+type Headers = SignatureHeaders & Partial<ContentHeaders>
 
 async function createClient(env: Env) {
```
sidvishnoi, line 38: Maybe also add a comment that it's based on extension code (along with link to code)

frontend/app/utils/open-payments.server.ts
```diff
@@ -174,3 +221,105 @@ export async function getWalletAddress(
...
+    .replace(/\s/g, '')
+
+  const derBytes = atob(base64Content)
+
+  const bytes = new Uint8Array(derBytes.length)
+  for (let i = 0; i < derBytes.length; i++) {
+    bytes[i] = derBytes.charCodeAt(i)
+  }
+
+  return bytes.slice(-32)
```
sidvishnoi, line 324: Maybe we can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64 this?

sidvishnoi: Or https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/from instead of for loop at least?

frontend/app/utils/open-payments.server.ts
DarianM: could not make it work, but something to replace `JSON.parse`  `JSON.stringify`

frontend/app/utils/open-payments.server.ts
raducristianpopa: Is they key stored as a PEM in CloudFlare? If so, I would suggest storing it as a Base64 string and just use `Buffer.from` when initializing the client. 

See: https://github.com/interledger/interledger-pay/blob/93079c16bc960464e13a93c3225df756b82e2cee/app/lib/open-payments.server.ts#L16

frontend/app/utils/open-payments.server.ts
```diff
@@ -1,20 +1,85 @@
...
+  'Content-Type': string
+}
+type Headers = SignatureHeaders & Partial<ContentHeaders>
 
+/**
+ * Creates an authenticated Open Payments client for making signed requests.
+ * Based on the Interledger Web Monetization Extension implementation:
+ * @see https://github.com/interledger/web-monetization-extension/blob/main/src/background/services/openPayments.ts#L163
+ */
 async function createClient(env: Env) {
```
raducristianpopa, line 43: Since this runs in CF Workers, maybe we can initialize the client as a singleton, and we would not create a new client when a worker is reused?

i.e.:

```ts
let client: AuthenticatedClient;

async function createClient() {
  if(client) return client
  
  return await createAuthenticatedClient({ ... })
}
```

frontend/app/utils/open-payments.server.ts
DarianM: 👍 suggestion works too in preview!

frontend/vite.config.ts
```diff
@@ -24,5 +24,14 @@ export default defineConfig({
...
     tsconfigPaths()
-  ]
+  ],
+  resolve: {
+    alias: {
+      crypto: 'crypto-browserify'
+    }
+  },
+  build: {
+    sourcemap: 'inline',
```
sidvishnoi, line 34: To avoid increasing size of bundled JS file (it'll use source map as needed).
```suggestion
    sourcemap: true,
```

https://developers.cloudflare.com/pages/functions/source-maps/

sidvishnoi (approved):

PR #112, opened 2025-05-21
fix: update URLs used by embed script

DarianM:
### Description

update `vite.config.ts` file in the `embed` directory to introduce environment-specific configurations


embed/vite.config.ts
```diff
@@ -5,7 +5,21 @@ export default defineConfig(({ mode }) => {
   const isDevelopment = mode === 'development'
 
   return {
-    envDir: resolve(__dirname),
+    define: {
+      'import.meta.env.VITE_SCRIPT_FRONTEND_URL': JSON.stringify(
+        isDevelopment
+          ? 'http://localhost:3000'
```
sidvishnoi, line 11: Might be needed how it's used.
```suggestion
          ? 'http://localhost:3000/'
```

embed/vite.config.ts
```diff
@@ -5,7 +5,21 @@ export default defineConfig(({ mode }) => {
...
+          : 'https://webmonetization.org.pages.dev/tools/'
+      ),
+      'import.meta.env.VITE_SCRIPT_API_URL': JSON.stringify(
+        isDevelopment
+          ? 'http://localhost:8787'
```
sidvishnoi, line 16: Similarly for this. We use string concatenation when building URLs, so need to be careful if we want our constants to have trailing `/` or not/
```suggestion
          ? 'http://localhost:8787/'
```

embed/.env.production
```diff
@@ -0,0 +1,3 @@
+VITE_SCRIPT_FRONTEND_URL=https://webmonetization.org/tools/
+VITE_SCRIPT_API_URL=https://tools-api.webmonetization.org/
```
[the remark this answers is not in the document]
DarianM: thought of that, we can, i don't want it to cause any confusion that wm.org has an api
but we also have cdn.webmonetization.org so it's good, let's edit.

PR #113, opened 2025-05-22
chore(frontend): use worker instead of pages

DarianM:
### Description

Transitioning frontend from Cloudflare Pages Functions to a Cloudflare Workers-based setup.
Cloudflare since last month started recommended workers over pages: https://developers.cloudflare.com/workers/static-assets/migration-guides/migrate-from-pages/


DarianM:
> Should check how this compares to the [Vite plugin approach from Cloudflare](https://github.com/remix-run/react-router-templates/blob/main/cloudflare).

Having troubles integrating the cloudflare/vite plugin. It  adds configuration complexity, duplicating config between remix, vite and cloudflare plugin.

the cloudflare/vite-plugin isn't that valuable when using a framework like remix.
Already have good cloudflare workers support via @remix-run/cloudflare

sidvishnoi (approved):

PR #127, opened 2025-05-28
fix(frontend): single fetch remix feature

### LINKED ISSUE #122: Bug: Back button in PageHeader component not functioning as expected
### Description
the `Back` button or the header link  PageHeader component do not properly redirect to homepage when clicked

Image

**Note:** reproducible only webmonetization.org/tools. Also, check network for error when reproducing this

DarianM:
### Description

Closes #122 

When using remix's `v3_singleFetch` feature flag with webmonetization.org/tools in production, single fetch data requests are made to incorrect URLs, resulting in 404 errors.

Expected Behavior
Single fetch data requests should respect the basename configuration and request:
`https://webmonetization.org/tools/tools.data`

Requested Changes
Added manual redirect handling in Cloudflare Worker.
Disabling the feature flag is also an option

Enabled feature flag context:
https://remix.run/docs/en/main/guides/single-fetch#single-fetch

### DIFF
--- a/frontend/worker.ts
+++ b/frontend/worker.ts
@@ -19,6 +19,15 @@ export default {
         return Response.redirect(new URL(`${APP_BASEPATH}/`, request.url), 302)
       }
 
+      // handle single fetch data loading and streaming format requests
+      // part of v3_singleFetch feature flag
+      if (url.pathname === '/tools.data') {
+        return Response.redirect(
+          new URL(`${APP_BASEPATH}/tools.data`, request.url),
+          301
+        )
+      }
+
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore it may not be built during type checking stage
       const build = await import('./build/server/index.js')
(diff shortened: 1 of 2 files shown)


frontend/worker.ts
```diff
@@ -19,6 +19,15 @@ export default {
         return Response.redirect(new URL(`${APP_BASEPATH}/`, request.url), 302)
       }
 
+      // handle single fetch data loading and streaming format requests
+      // part of v3_singleFetch feature flag
+      if (url.pathname === '/tools.data') {
```
sidvishnoi, line 24: This is the only route with such issue? Do we have an issue open in remix repo? Or maybe it was fixed in newer versions?

sidvishnoi (approved):
If it works, it works.

DarianM:
> If it works, it works.

appears to be solved with ReactRouter `7.3.0`
https://github.com/remix-run/react-router/issues/12295#issuecomment-2669440359

PR #131, opened 2025-06-09
fix(frontend): dynamic tool path

DarianM:
### Description

- Include the tool type in paths to provide support for configuration options for other tools.
- fix `link` path property in presets for widget & button


frontend/app/components/PageHeader.tsx
```diff
@@ -8,14 +8,12 @@ import { Chevron } from './icons.js'
 export const PageHeader = ({
   elementType,
   title,
-  contentOnlyLink,
```
sidvishnoi, line 11: Can we move changes to `contentOnlyLink` to a separate PR?

frontend/app/lib/presets.ts
```diff
@@ -139,7 +139,7 @@ export const availableTools = [
     title: 'Widget',
     image: widgetImg,
     bgColor: 'from-wm-red to-wm-red-fade',
-    link: 'create/widget',
+    link: 'widget',
```
sidvishnoi, line 142: I'm a little confused. Even before this PR, https://webmonetization.org/tools/banner works, and it isn't https://webmonetization.org/tools/create/banner. Why does the current thing work even?

frontend/app/components/PageHeader.tsx
DarianM: yes we can 👍

frontend/app/lib/presets.ts
sidvishnoi: Yes, why does https://webmonetization.org/tools/banner work already (even before this PR), including the links from home page (https://webmonetization.org/tools/)

PR #143, opened 2025-06-12
fix(embed): banner css file path

DarianM:
### Requested changes

- Fix the css file request path for the embed script.
Before: `webmonetization.org/tools/tools/css/banner.css`
After: `webmonetization.org/tools/css/banner.css`

- Improvement. Wait for the css file to load inside the shadow root before displaying the banner. Noticed an issue on localhost where the banner was displaying in the root before the CSS loaded.


sidvishnoi:
We don't need to await css load to act further on the element. Is there a reason to wait? And why not include CSS within element like we do for the widget?

PR #144, opened 2025-06-13
fix(api): log errors

DarianM:
### Requested changes

Current api logs are not very useful

### DIFF
--- a/api/src/index.ts
+++ b/api/src/index.ts
@@ -12,6 +12,7 @@ import {
   WalletAddressParamSchema
 } from './schemas/payment.js'
 import type { ConfigVersions } from './types.js'
+import { createHTTPException, serializeError } from './utils/utils.js'
 
 export type Env = {
   AWS_ACCESS_KEY_ID: string
@@ -61,13 +62,12 @@ app.use('*', async (c, next) => {
       )
     }
 
-    console.error('Unexpected error:', error)
+    const serializedError = serializeError(error)
+    console.error('Unexpected error: ', serializedError)
     return c.json(
       {
         error: {
-          message:
-            error instanceof Error ? error.message : 'Internal server error',
-          details: error instanceof Error ? error.stack : undefined,
+          ...serializedError,
           code: 'INTERNAL_ERROR'
         }
       },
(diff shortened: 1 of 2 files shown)


api/src/index.ts
```diff
@@ -60,11 +60,14 @@ app.use('*', async (c, next) => {
...
     }
+
+    console.error('Unexpected error:', error)
     return c.json(
       {
         error: {
           message:
             error instanceof Error ? error.message : 'Internal server error',
+          details: error instanceof Error ? error.stack : undefined,
           code: 'INTERNAL_ERROR'
```
sidvishnoi, line 71:
```suggestion
          code: 'INTERNAL_ERROR',
          ...JSON.parse(JSON.stringify(error, Object.getOwnPropertyNames(error)))
```

api/src/index.ts
```diff
@@ -120,14 +123,18 @@ app.post(
 
       return json(result)
     } catch (error) {
+      console.error('Payment quote creation error: ', {
+        error: error instanceof Error ? error.message : String(error),
+        stack: error instanceof Error ? error.stack : undefined
+      })
```
sidvishnoi, line 129:
```suggestion
      console.error('Payment quote creation error: ', JSON.parse(JSON.stringify(error, Object.getOwnPropertyNames(error))))
```

sidvishnoi: Can extract this into a helper: `errorToJson`. This will get us all properties for that error, including `code` if it exists.

api/src/index.ts
```diff
@@ -61,13 +62,12 @@ app.use('*', async (c, next) => {
...
+    const serializedError = serializeError(error)
+    console.error('Unexpected error: ', serializedError)
     return c.json(
       {
         error: {
-          message:
-            error instanceof Error ? error.message : 'Internal server error',
-          details: error instanceof Error ? error.stack : undefined,
+          ...serializedError,
           code: 'INTERNAL_ERROR'
```
sidvishnoi, line 71: We're overriding error code from `error` here. Intentional?

DarianM: it's a mistake. let's give `code` if it exists from that error, as you mentioned before

PR #160, opened 2025-07-12
feat: redesign revshare

adrianboros:


frontend/app/assets/svg.tsx
```diff
@@ -291,3 +291,42 @@ export const SVGArrow1 = ({ className }: { className: string }) => (
...
+      fill="white"
+    />
+  </svg>
+)
+export const SVGDeleteScript = ({
+  className = '',
+  width = 20,
```
DarianM, line 312: let's use the `SvgIcon` type for the svg's from now on, to make the className  field required.
let's move the `width` and `height` directly at the call site where the svg is used, to be consistent with the rest of the svgs

frontend/app/components/redesign/components/Header.tsx
```diff
@@ -47,7 +47,7 @@ export const Header = () => {
           className="relative flex items-center justify-between bg-white shadow-[0px_12px_20px_0px_rgba(0,0,0,0.06)] rounded-lg mx-4 mt-4 h-14 py-1 pl-4 pr-2 md:max-w-7xl md:mx-auto md:px-xl md:py-xs md:mt-0 md:rounded-none md:shadow-none md:bg-transparent md:h-auto md:w-full"
         >
           <RemixNavLink
-            to="/"
+            to="https://webmonetization.org/"
```
DarianM, line 50: @kjmitchelljr @sidvishnoi  I do not understand how `to="/"` does not redirect to homepage already. using "/" looks just fine to me

DarianM (approved):
implementation respects the original WICG implementation. 🎉

frontend/app/components/redesign/components/revshare/ImportTagModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+  onClose?: () => void
+  onConfirm?: () => void
+  tag: string
+  errorMessage: string
+  setTag: (tag: string) => void
+  setImportError: (error: string) => void
+  className?: string
+}
+
+export const ImportTagModal: React.FC<ImportTagModalProps> = ({
```
sidvishnoi, line 18: When this modal is opened, focus on the input field.

sidvishnoi, line 18: Should do as follow up: trap focus within modal while it's open.

frontend/app/components/redesign/components/revshare/ImportTagModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+        >
+          <BodyStandard>Import existing revshare configuration</BodyStandard>
+          <textarea
+            id="linkTagInput"
+            className={cx(
+              'py-sm pl-md pr-xs rounded-lg border border-silver-300 resize-none',
+              'w-full max-w-full h-[136px]',
+              'focus:border-field-border-focus focus:outline-none focus:ring-1 focus:ring-primary-focus'
+            )}
+            placeholder="Your link tag here"
```
sidvishnoi, line 94: Consider adding a placeholder with `<link ...>` instead of this.

frontend/app/components/redesign/components/revshare/ImportTagModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <SVGClose className="w-6 h-6" />
+        </button>
+        <div
+          className={cx(
+            'flex flex-col items-center justify-center px-md',
+            errorMessage ? '' : 'gap-lg'
+          )}
+        >
+          <BodyStandard>Import existing revshare configuration</BodyStandard>
+          <textarea
```
sidvishnoi, line 87: This textarea needs a label.

frontend/app/components/redesign/components/revshare/ShareInput.tsx
sidvishnoi, line 1: Should work on the accessibility of this (ok as follow up). The table was accessible, but hard to make responsive. We did the responsive part here, but sabotaged its accessibility.

frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -0,0 +1,133 @@
...
+            step="any"
+            value={typeof percent === 'number' ? Math.round(percent * 100) : ''}
+            onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+              onChangePercent(Number(e.target.value) / 100)
+            }
+            disabled={percentDisabled}
+          />
+        </div>
+        <div className="hidden md:block">
+          <ToolsSecondaryButton
```
sidvishnoi, line 121: This button needs a label.

frontend/app/components/redesign/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,28 @@
+import { useState, useEffect, useCallback } from 'react'
+
+export const useCopyToClipboard = (link: string) => {
```
sidvishnoi, line 3: This function can be generic enough:
```suggestion
export const useCopyToClipboard = (text: string) => {
```

and pass it full `<link...>` as input.

frontend/app/lib/revshare.ts
```diff
@@ -0,0 +1,381 @@
...
+ * @property {string} pointer - The payment pointer of the recipient.
+ * @property {number} [weight] - The numerical weight of the share, used to calculate the distribution.
+ * @property {number} [percent] - The calculated percentage of the total revenue for this share.
+ */
+export type Share = {
+  name?: string
+  pointer: string
+  weight?: number
+  percent?: number
+}
```
sidvishnoi, line 29: Given we're using typescript, we should probably do it like this:
```ts
/** Represents a single revenue share participant. */
export type Share {
  /** An optional name for the recipient for display purposes. */
  name?: string
}
```

frontend/app/routes/prob-revshare.tsx
```diff
@@ -0,0 +1,272 @@
...
+              >
+                Add rev share
+              </ToolsPrimaryButton>
+            </div>
+          </div>
+        </Card>
+        <div className="my-lg md:my-md">
+          <RevShareChart shares={shares} />
+        </div>
+        <div className="flex flex-col gap-md mb-2xl">
```
sidvishnoi, line 215: Consider extracting this into a separate component, as it doesn't need to re-render on any state change.

frontend/app/components/redesign/components/revshare/ImportTagModal.tsx
kjmitchelljr: Will create a ticket for this in board

frontend/app/components/redesign/components/revshare/ShareInput.tsx
kjmitchelljr: Will create a ticket for this in board

frontend/app/components/redesign/components/revshare/ShareInput.tsx
sidvishnoi: Label as in any text content for the button - either `aria-label` or some `.sr-only` text content.

frontend/app/components/redesign/components/revshare/ShareInput.tsx
DarianM: [#193](https://github.com/interledger/publisher-tools/issues/193) same as above 🫡

Issue #198, opened 2025-07-28
[BUG] Link Tag Generator Mobile text issue


Tymmmy:
### Steps to reproduce

1. Go to Publisher Tools -> Link Tag generator MOBILE screen
3. Scroll down to the copyright text
4. Observe that for very small sizes the copyright text is not centered and for almost all Iphone screen sizes the text is very close to the above text.

### Expected result

The copyright text is centered and also at the bottom of the page

### Actual result

The copyright text on Mobile screen sizes is not always correct

### Screenshots or videos

<img width="487" height="792" alt="Image" src="https://github.com/user-attachments/assets/737f3efc-726a-405d-9306-18bc95f23fc1" />

<img width="407" height="783" alt="Image" src="https://github.com/user-attachments/assets/b3d73740-4a65-41d7-8b84-97acfa93b332" />

### Affected component

Link Tag

### Device type

Mobile

### Browsers

Safari

PR #210, opened 2025-07-30
fix(redesign): footer spacing and text misalignment

kjmitchelljr:
Resolves issue #198 for the Copyright text on mobile.


DarianM:
@kjmitchelljr we had to add padding to the footer due to fixed call to action buttons in mobile view for `/banner`.

Adding padding to `footer` by me was a mistake as it affected all routes.
I moved the padding to the banner tool path only.
Please check the last commit for any inconvenience.

Otherwise LGTM

kjmitchelljr:
Looks good to me 🫡 @DarianM

PR #214, opened 2025-08-04
feat: create /widget route and widget builder

DarianM:
## Added Changes

`/widget.tsx` - new widget builder route with preview functionality

### shared types/utils

`shared/utils/index.ts` - Enhanced type definitions for widget/banner positions and configurations
`shared/utils/index.ts` - Added shared utility functions
`shared/default-data/index.ts` - Updated default configurations for widget support

### Frontend components

`TabSelector.tsx`
- Previously, the navigation with Arrow keys (ArrowLeft and ArrowRight) would always run. Now, the whole Arrow key logic only runs if you are not editing (!editingId).
This prevents keyboard navigation between tabs when a tab is currently being edited, avoiding potential focus and state bugs.
- Fix: Switched the input from a controlled component (using value) to an uncontrolled component (using defaultValue). This avoids issues where the input's value might get reset or not update as expected while editing

### Routing & Validation

validate.server.ts - enhanced validation for widget configurations
api.config.$type.ts - updated API handling for widget configurations

###  Summary

- Unified Builders : reusable components that work for both banner and widget tools
- State management: store to handle multiple tool types
- Component re-usability:  more generic and reusable
- Position Independence: ensured widget and banner positioning logic is completely separate
- Path Tracking:  hook for tracking user navigation and behavior

### DIFF
--- a/shared/utils/index.ts
+++ b/shared/utils/index.ts
@@ -32,7 +32,7 @@ export async function getWalletAddress(
     throw new WalletAddressFormatError('Invalid wallet address format')
   }
 
-  return normalizeWalletAddress(json)
+  return json
 }
 
 export function isWalletAddress(
@@ -56,9 +56,7 @@ export function toWalletAddressUrl(s: string): string {
   return s.startsWith('$') ? s.replace('$', 'https://') : s
 }
 
-export function normalizeWalletAddress(
-  walletAddress: WalletAddress
-): WalletAddress {
+export function normalizeWalletAddress(walletAddress: WalletAddress): string {
   const IS_INTERLEDGER_CARDS =
     walletAddress.authServer === 'https://auth.interledger.cards'
   const url = new URL(toWalletAddressUrl(walletAddress.id))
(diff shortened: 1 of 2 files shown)


frontend/app/utils/validate.server.ts
```diff
@@ -1,7 +1,11 @@
...
-import { WalletAddressFormatError } from '~/lib/types.js'
-import { isWalletAddress, toWalletAddressUrl } from '@shared/utils'
+import {
+  checkHrefFormat,
+  isValidWalletAddress,
+  toWalletAddressUrl,
+  WalletAddressFormatError
+} from '@shared/utils'
 
 const rangeError = { message: 'Value has to be between 16 and 24' }
```
sidvishnoi, line 10: Don't forget to update this message (10-30) now.

DarianM: 👍 still debating if we should add different mix-max font values sizes for banner and widget

frontend/app/routes/widget.tsx
```diff
@@ -0,0 +1,502 @@
...
+            </div>
+          </div>
+        </div>
+      )}
+
+      {snap.modal?.type === 'save-error' && (
+        <div className="fixed inset-0 bg-slate-500 bg-opacity-75 transition-opacity z-50">
+          <div className="fixed inset-0 z-50 overflow-y-auto">
+            <div className="flex min-h-full items-center justify-center p-4 text-center">
+              <SaveResultModal
```
sidvishnoi, line 448: Consider moving all these modals into a separate component. Then the `Widget` will be smaller and easier to scan.

shared/utils/index.ts
```diff
@@ -1,30 +1,38 @@
...
+      {
+        cause: error
+      }
+    )
   }
+  if (!isWalletAddress(json)) {
+    throw new WalletAddressFormatError('Invalid wallet address format')
+  }
+
+  return normalizeWalletAddress(json)
```
sidvishnoi, line 35: This modification might be too much to do it here. We won't know if we had `ilp.dev` address this way.

PR #220, opened 2025-08-06
fix(frontend): use `font-mono` on code blocks

sidvishnoi:
Font family approved by Radu M.

| Before | After |
| --- | --- |
| image | |


DarianM (approved):
very nice. much  clearer

PR #234, opened 2025-08-07
fix(frontend): improve a11y for footer

DarianM:
Follow up of #230 ???

### DIFF
--- a/frontend/app/components/redesign/components/footer/EnhancedFooter.tsx
+++ b/frontend/app/components/redesign/components/footer/EnhancedFooter.tsx
@@ -37,13 +37,13 @@ const socialLinks = [
 ]
 
 const SocialLinks = ({ className }: { className: string }) => (
-  <div className={className}>
-    {socialLinks.map((social, index) => (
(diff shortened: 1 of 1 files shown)


sidvishnoi (commented):
Will review in detail from the computer tomorrow.

frontend/app/components/redesign/components/footer/EnhancedFooter.tsx
```diff
@@ -95,27 +98,39 @@ export const EnhancedFooter = () => {
...
+      <div
+        className="hidden md:block w-full h-px bg-purple-100"
+        aria-hidden="true"
+      />
 
-      <div className="hidden md:flex flex-row gap-sm items-center justify-start">
+      <section
+        className="hidden md:flex flex-row gap-sm items-center justify-start"
+        aria-label="Branding"
+      >
```
sidvishnoi, line 125: It's ok without separate section. Having too many sections is confusing for users who browse via landmarks. Use section for where it makes sense for it to be announced separately.

sidvishnoi (approved):

Issue #223, opened 2025-08-06
[IMPROVEMENT] Percentage Field readonly for Rev Share Generator


kjmitchelljr:
### Describe the feature you would like to request

Update the revenue share generator table to where the Percentage field for revenue share split is read only and that the only thing the user changes is the Weight to determine the split configuration.

### Describe the solution you would like

Percentage Field becomes readonly for Revenue Share Generator

### Additional context

Simplifies user experience for selecting the revenue split between recipients and fixes UI issue found in with Percentage selector.

PR #212, opened 2025-07-31
fix(accessibility): added accessibility features for rev share generator

kjmitchelljr:
Resolves https://github.com/interledger/publisher-tools/issues/193 by adding suggested accessibility features for the Revenue Share Generator Table having tested the following items

- [ ] Screen readers can navigate and understand the data structure
- [x] All input fields have proper accessible names
- [x] Column headers are associated with their data
- [x] Validation errors are announced to screen readers
- [ ] Keyboard navigation works smoothly
- [x] Component maintains responsive design

PR #235, opened 2025-08-07
fix(/prob-revshare)!: make percentage field as read only

kjmitchelljr:
Closes https://github.com/interledger/publisher-tools/issues/223
Changing the percentage field of the revenue share generator to be read only and not an input that is editable by the user

### DIFF
--- a/frontend/app/components/redesign/components/revshare/ShareInput.tsx
+++ b/frontend/app/components/redesign/components/revshare/ShareInput.tsx
@@ -219,7 +219,10 @@ export const ShareInput = React.memo(
           </div>
         </div>
         <div role="cell" aria-labelledby="col-percentage">
-          <div id={percentInputId} className="ml-2xs md:ml-0 md:text-center">
+          <div
+            id={percentInputId}
+            className="ml-2xs md:ml-0 md:text-center text-field-helpertext-default"
+          >
             {Math.round(percent * 100)}%
           </div>
         </div>


frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -11,34 +11,83 @@ interface ShareInputProps {
...
+  children: React.ReactNode
+}) => {
+  return (
+    <div
+      role="table"
+      aria-labelledby="revshare-table-caption"
+      className="contents"
+    >
+      <div id="revshare-table-caption" className="sr-only">
+        Revenue sharing recipients table
```
sidvishnoi, line 37:
```suggestion
        Revenue sharing recipients
```

frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -11,34 +11,83 @@ interface ShareInputProps {
...
 }
 
 const GRID_COLS = 'md:grid-cols-[16rem_1fr_6rem_6rem_auto]'
 const GRID_GAP = 'md:gap-x-md'
 
+export const ShareInputTable = ({
+  children
+}: {
+  children: React.ReactNode
+}) => {
```
sidvishnoi, line 29:
```suggestion
export const ShareInputTable = ({ children }: React.PropsWithChildren) => {
```

DarianM:
> @DarianM @sidvishnoi preview doesn't seem to be working, I did branch this from latest main and then merged in my changes in `193-bug-revenue-share-generator-accesability-part-2`

we need to merge #212 first, to not review the same changes twice 🫡

frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -11,34 +11,83 @@ interface ShareInputProps {
...
+  return (
+    <div
+      role="table"
+      aria-labelledby="revshare-table-caption"
+      className="contents"
+    >
+      <div id="revshare-table-caption" className="sr-only">
```
sidvishnoi, line 36:
```suggestion
      <div id="revshare-table-caption" role="caption" className="sr-only">
```

frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -11,34 +11,83 @@ interface ShareInputProps {
...
+        Revenue sharing recipients table
+      </div>
+      {children}
+    </div>
+  )
+}
+
 export const ShareInputHeader = () => {
```
sidvishnoi, line 44: The header should be inside a `role="rowgroup"` (acts as `<thead>`), and then all the data inside another `role="rowgroup"` (as it already is, acts as `<tbody>`).

frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -112,26 +212,33 @@ export const ShareInput = React.memo(
...
-            disabled={percentDisabled}
-          />
+        <div role="cell" aria-labelledby="col-percentage">
+          <div
+            id={percentInputId}
+            aria-label={`Calculated percentage: ${Math.round(percent * 100)}%`}
+            className="w-full px-sm py-xs rounded-sm border border-field-border"
+          >
+            {Math.round(percent * 100)}%
+          </div>
```
sidvishnoi, line 230: This needs some styling update so that it feels like readonly.

frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -224,26 +218,13 @@ export const ShareInput = React.memo(
...
-            onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
-              onChangePercent(Number(e.target.value) / 100)
-            }
-            disabled={percentDisabled}
-            aria-disabled={percentDisabled}
-            aria-describedby={`percent-description-${index}`}
-          />
-          <div id={`percent-description-${index}`} className="sr-only">
-            Calculated from weight, or can be edited directly.
+            aria-label={`Calculated percentage: ${Math.round(percent * 100)}%`}
```
sidvishnoi, line 224: As this div has no assigned role, `aria-label` will not be used (except when using VoiceOver).
https://benmyers.dev/blog/dont-use-aria-label-on-static-text-elements/

Consider using `sr-only` and `aria-hidden` if you really want to show "calculated percentage: {}", which I think is redundant, as we're in a table and the cell has a header.

sidvishnoi: What I mean is, there's an aria-label on a plain div here, which isn't right.

sidvishnoi:
I mean only the styles for percentage value. Maybe make it a little grayer so it looks read-only.

kjmitchelljr:
@sidvishnoi should be good now, though not sure if we'll put this PR on hold until there's a final conclusion on how we'll handle weight/percentage fields for rev share, thoughts?

PR #236, opened 2025-08-08
fix(LinkTagGenerator): improve wallet adress validation

adrianboros:


shared/utils/index.ts
```diff
@@ -112,3 +112,33 @@ export function checkHrefFormat(href: string): string {
 
   return href
 }
+
+export const isValidPointer = (input: string): string | false => {
```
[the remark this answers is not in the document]
sidvishnoi: That way, we'll have a function, `validatePointer` which will either return a normalized wallet address, or throw with error message.

kjmitchelljr:
Thank you for getting started on this work @adrianboros, also note the same validation will be used for the Rev Share Generator wallet addresses. Then to Sid's point Alex did mention on a call that there is a way to validate the existence of an address, but I believe that would involve some type of request going and getting that response

shared/utils/index.ts
```diff
@@ -113,32 +113,92 @@ export function checkHrefFormat(href: string): string {
...
-      urlString = input.replace(/^\$/, 'https://')
+export class WalletValidationError extends Error {
+  constructor(message: string, options?: { cause?: unknown }) {
+    super(message)
+    this.name = 'WalletValidationError'
+    if (options?.cause !== undefined) {
+      // @ts-expect-error: Error.cause is available in modern TS/Node; keep for browsers too
+      this.cause = options.cause
     }
+  }
```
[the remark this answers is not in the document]
adrianboros: Thanks, 
I pushed suggestion from local as it's missing the closing } in the above suggestion

shared/utils/index.ts
```diff
@@ -112,3 +112,89 @@ export function checkHrefFormat(href: string): string {
...
+    throw new WalletValidationError('domain name is not valid')
+  }
+
+  const path = url.pathname === '/' ? '/.well-known/pay' : url.pathname
+  return `${url.origin}${path}`
+}
+
+export async function validateAndConfirmPointer(url: string): Promise<string> {
+  const validUrl = validateWalletAddressOrPointer(url)
+  await confirmWalletAddress(validUrl)
```
sidvishnoi, line 198: I think the error coming from fetch should not stop from seeing the generated link tag - consider temporary network issues, CORS errors (particularly on tools website origin), or users just trying out the tool.

We can still show the error, just I think we shouldn't stop user from seeing the link tag in case of network error (or even 404).

shared/utils/index.ts
```diff
@@ -112,3 +112,89 @@ export function checkHrefFormat(href: string): string {
...
+
+  if (!isWalletAddress(wallet)) {
+    throw new WalletValidationError('does not have a valid wallet response')
+  }
+}
+
+export const validateWalletAddressOrPointer = (input: string): string => {
+  if (!input || typeof input !== 'string') {
+    throw new WalletValidationError('must be a string')
+  }
```
sidvishnoi, line 146: Users might not understand what this error means. Maybe saying this field is required is better (or we can even mark the input as `required` so we get native form validation).

image

frontend/app/components/redesign/components/LinkTagGenerator.tsx
```diff
@@ -94,7 +84,9 @@ export const LinkTagGenerator = () => {
           onChange={(e) => handleOnChange(e)}
           error={
             invalidUrl
-              ? 'The payment pointer you have entered is not valid or cannot be found'
+              ? error
+                ? error
+                : 'The payment pointer you have entered is not valid or cannot be found'
```
DarianM, line 89: i do not understand this change

shared/utils/index.ts
```diff
@@ -112,3 +112,89 @@ export function checkHrefFormat(href: string): string {
...
+  walletAddress: string
+): Promise<void> {
+  const wallet = await fetchWalletDetails(walletAddress)
+
+  if (!isWalletAddress(wallet)) {
+    throw new WalletValidationError('does not have a valid wallet response')
+  }
+}
+
+export const validateWalletAddressOrPointer = (input: string): string => {
```
DarianM, line 143: i believe there's a lot of duplicate code with what we already have in other utility functions.
i think we can reuse some

frontend/app/components/redesign/components/InputField.tsx
```diff
@@ -38,7 +42,9 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
           name={fieldId}
           {...props}
         />
-        {error && <p className="text-xs text-text-error">{error}</p>}
```
DarianM, line 41: add additional wrapper  to prevent layout shift when the error message appears

shared/utils/index.ts
adrianboros: error messages are tailored for the link tag generator - see image - and it follows this structure: 

1. Your Payment Pointer
2. red border wallet address field
3. error message

Screenshot 2025-08-12 at 16 25 29

frontend/app/components/redesign/components/LinkTagGenerator.tsx
```diff
@@ -37,24 +21,33 @@ export const LinkTagGenerator = () => {
...
       setInvalidUrl(false)
+      setError('')
 
-      const validatedPointer = isValidPointer(pointerInput)
+      setParsedLinkTag(htmlEncodePointer(pointerInput))
 
-      if (validatedPointer) {
-        setParsedLinkTag(htmlEncodePointer(validatedPointer))
+      try {
+        await validateWalletAddressOrPointer(pointerInput)
```
sidvishnoi, line 39: Something changed here, and now my payment pointer gives this:
image

DarianM: brought back the previous validations

shared/utils/index.ts
```diff
@@ -112,3 +112,89 @@ export function checkHrefFormat(href: string): string {
...
+    this.name = 'WalletValidationError'
+  }
+}
+
+export async function fetchWalletDetails(url: string): Promise<WalletAddress> {
+  const res = await fetch(url, {
+    headers: { Accept: 'application/json' }
+  })
+  if (!res.ok) {
+    throw new WalletValidationError('is not a valid wallet address')
```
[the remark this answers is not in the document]
DarianM: how about adding a spinner on the button until validation/confirmation is done?

frontend/app/components/redesign/components/LinkTagGenerator.tsx
```diff
@@ -37,24 +21,29 @@ export const LinkTagGenerator = () => {
...
+    async (e: React.FormEvent) => {
       e.preventDefault()
       setInvalidUrl(false)
+      setError('')
 
-      const validatedPointer = isValidPointer(pointerInput)
-
-      if (validatedPointer) {
+      try {
+        const validatedPointer = await validateAndConfirmPointer(pointerInput)
```
sidvishnoi, line 37: Might want a spinner on button as wallet address fetching can take time sometimes.

shared/utils/index.ts
```diff
@@ -112,3 +112,87 @@ export function checkHrefFormat(href: string): string {
...
   return href
 }
+
+export async function fetchWalletDetails(url: string): Promise<WalletAddress> {
+  const res = await fetch(url, {
+    headers: { Accept: 'application/json' }
+  })
+  if (!res.ok) {
+    if (res.status === 404) {
+      throw new WalletAddressFormatError('this wallet address does not exist')
```
sidvishnoi, line 122: Consistent with other error messages.
```suggestion
      throw new WalletAddressFormatError('does not exist')
```

shared/utils/index.ts
```diff
@@ -112,3 +112,87 @@ export function checkHrefFormat(href: string): string {
...
+  const wallet = await fetchWalletDetails(walletAddress)
+
+  if (!isWalletAddress(wallet)) {
+    throw new WalletAddressFormatError('does not have a valid wallet response')
+  }
+}
+
+export const validateWalletAddressOrPointer = (input: string): string => {
+  if (!input || typeof input !== 'string') {
+    throw new WalletAddressFormatError('must be a string')
```
sidvishnoi, line 141:
```suggestion
    throw new WalletAddressFormatError('is required')
```

PR #238, opened 2025-08-11
fix(/prob-revshare): non-empty table

### LINKED ISSUE #219: [BUG]: rev-share empty table can be confusing
### Steps to reproduce

1. Go to rev share tool page
2. Keep Clicking on "delete" buttons

### Expected result

If I remove all rows, an empty row should be added automatically, instead of showing an empty table and expecting user will know they've to click "Add recipient" button.

### Actual result

The table becomes empty, confusing user.

### Screenshots or videos

Image

### Additional context

_No response_

### Affected component

Revenue Share

### Device type

Desktop

### Browsers

Firefox

### Browser version

_No response_

kjmitchelljr:
Resolves #219 for when the table is empty. Automatically will add the default of two rows to the table to better enhance the user experience

### DIFF
--- a/frontend/app/routes/prob-revshare.tsx
+++ b/frontend/app/routes/prob-revshare.tsx
@@ -145,7 +145,8 @@ function Revshare() {
 
   const handleRemove = useCallback(
     (index: number) => {
-      setShares(dropIndex(shares, index))
+      const newShares = dropIndex(shares, index)
+      setShares(newShares.length > 0 ? newShares : [newShare(), newShare()])
     },
     [shares, setShares]
   )


frontend/app/routes/prob-revshare.tsx
```diff
@@ -145,7 +145,8 @@ function Revshare() {
 
   const handleRemove = useCallback(
     (index: number) => {
-      setShares(dropIndex(shares, index))
+      const newShares = dropIndex(shares, index)
+      setShares(newShares.length > 0 ? newShares : [newShare(), newShare()])
```
sidvishnoi, line 149: Why not _maintain_ at least 2 all the time? Why wait for empty? If there's one left, can we still have an empty one automatically?

DarianM: Should we do in this PR a minimum of two distributions added for a generated distribution?

frontend/app/components/redesign/components/ToolsSecondaryButton.tsx
```diff
@@ -32,6 +32,7 @@ export function ToolsSecondaryButton({
         'gap-2',
         'focus:outline-none focus:ring-2 focus:ring-primary-focus',
         'transition-all duration-200',
+        'disabled:cursor-not-allowed disabled:hover:bg-transparent',
```
kjmitchelljr, line 35: Wondering if I should add some opacity here if it's disabled or is `cursor-not-allowed` sufficient?

PR #246, opened 2025-08-14
fix(frontend): tab selector horizontal scroll

DarianM:


sidvishnoi (changes requested):
I'd encourage getting a review from Radu/Bibi first before we let this in.

Issue #181, opened 2025-07-22
[BUG] UI  RevShare  |  align scrollbar with border


tselit:
### Steps to reproduce

1. Used this preview link: https://d07eb250-publisher-tools.webmonetization.workers.dev/tools/prob-revshare/
2. Add a test wallet address

### Expected result

I think the scrollbar of text area displaying the script would ideally be right and left aligned with the borders of the text area.

### Actual result

The scrollbar seems to start a few millimetres after the left border of the text area, and it ends a few millimetres before the border on the right side.

### Screenshots or videos

<img width=60% height="auto" alt="scroll bar not aligned to text area borders" src="https://github.com/user-attachments/assets/d1243609-b5fe-44d9-b919-8b6333941638" />

### Affected component

Revenue share

### Device type

Desktop

### Browsers

Chrome

### Browser version

Version 137.0.7151.119 (Official Build) (64-bit)

PR #248, opened 2025-08-16
fix(/prob-revshare): generated code block alignments

kjmitchelljr:
Addresses https://github.com/interledger/publisher-tools/issues/181.

### Changes

- refactored the `CodeBlock` component to be fully self-contained , improving reusability 
- center copy code button 
- scroll not overlapping/clipping the text


frontend/app/components/redesign/hooks/useCopyToClipboard.ts
```diff
@@ -1,12 +1,17 @@
 import { useState, useEffect, useCallback } from 'react'
 
-export const useCopyToClipboard = (text: string) => {
+export const useCopyToClipboard = (
+  text: string,
+  onCopy?: (copied: boolean) => void
```
sidvishnoi, line 5: `onCopy` doesn't seem the right name here. As the name suggests, `onCopy` should be fired only when it's actually copied. We're trying to emit state to parent here. Maybe renaming it to `callback` might be ok, as it won't cause this misunderstanding.

sidvishnoi (commented):
Some code quality comments. Rest looks good.

PR #251, opened 2025-08-19
fix(/prob-revshare): wallet address validation

### LINKED ISSUE #188: [BUG] RevShare wallet addresses/payment pointers are not validated before generating link
### Steps to reproduce

1. Go to https://86fa4fc5-publisher-tools.webmonetization.workers.dev/tools/prob-revshare/
2. Enter an payment pointer that does not exist, but starts with $ OR https://

### Expected result

- The payment pointer should be marked as invalid and flagged accordingly

### Actual result

- The link is generated and payment pointer is considered valid

### Screenshots or videos

Image

### Additional context

_No response_

### Affected component

Revenue Share

### Device type

Desktop

### Browsers

Chrome

### Browser version

_No response_

kjmitchelljr:
Resolves #188 to validate wallet addresses before generating link tag and pie chart

### DIFF
--- a/frontend/app/components/redesign/hooks/useDebounceValidation.ts
+++ b/frontend/app/components/redesign/hooks/useDebounceValidation.ts
@@ -10,7 +10,10 @@ interface ValidationState {
   error: string | null
 }
 
-export function useDebounceValidation(value: string, delay: number = 500): ValidationState {
+export function useDebounceValidation(
+  value: string,
+  delay: number = 500
+): ValidationState {
   const [validationState, setValidationState] = useState<ValidationState>({
     isValidating: false,
     isValid: null,
@@ -30,13 +33,13 @@ export function useDebounceValidation(value: string, delay: number = 500): Valid
         return
       }
 
-    setValidationState({
-      isValidating: true,
-      isValid: null,
-      error: null
-    })
+      setValidationState({
+        isValidating: true,
+        isValid: null,
+        error: null
+      })
 
-      let finalState: Pick<ValidationState, 'isValid' | 'error'>;
+      let finalState: Pick<ValidationState, 'isValid' | 'error'>
 
       try {
         await validateAndConfirmPointer(pointer)
--- a/frontend/app/components/redesign/hooks/useDebounceValidation.ts
+++ b/frontend/app/components/redesign/hooks/useDebounceValidation.ts
@@ -32,11 +32,11 @@ export function useDebounceValidation(
       return
     }
 
-    setValidationState({
-      isValidating: true,
+    setValidationState(prevState => ({
+      ...prevState,
       isValid: null,
       error: null
-    })
+    }))
 
     const validatePointer = async (pointer: string) => {
       let finalState: Pick<ValidationState, 'isValid' | 'error'>
@@ -64,6 +64,10 @@ export function useDebounceValidation(
     }
 
     const timeoutId = setTimeout(() => {
+      setValidationState(prevState => ({
+        ...prevState,
+        isValidating: true
+      }))
       validatePointer(value)
     }, delay)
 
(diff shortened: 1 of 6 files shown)


frontend/app/routes/prob-revshare.tsx
```diff
@@ -46,24 +46,15 @@ export default function RevsharePageWrapper() {
...
-}
-
 function Revshare() {
   const navigate = useNavigate()
   const [isModalOpen, setIsModalOpen] = useState(false)
   const [importTag, setImportTag] = useState('')
   const [importError, setImportError] = useState('')
   const { shares, setShares } = useShares()
+  const [validityMap, setValidityMap] = useState<
+    Record<number, boolean | null>
```
DarianM, line 56: i see no benefit in using `boolean | null`, let's leave only `boolean` as type

frontend/app/components/redesign/hooks/useDebounceValidation.ts
```diff
@@ -0,0 +1,81 @@
+import { useEffect, useState } from 'react'
+import {
+  validateAndConfirmPointer,
+  WalletAddressFormatError
+} from '@shared/utils/index'
+
+interface ValidationState {
+  isValidating: boolean
+  isValid: boolean | null
```
DarianM, line 9: here too, only `boolean`

frontend/app/components/redesign/hooks/useDebounceValidation.ts
```diff
@@ -0,0 +1,81 @@
...
+        isValidating: false,
+        isValid: null,
+        error: null
+      })
+      return
+    }
+
+    setValidationState(prevState => ({
+      ...prevState,
+      isValid: null,
```
DarianM, line 37: then this becomes `false` instead of `null`

frontend/app/routes/prob-revshare.tsx
```diff
@@ -133,7 +124,26 @@ function Revshare() {
...
+  )
+
+  const hasValidShares = useMemo(() => {
+    if (
+      shares.length === 0 ||
+      shares.length !== Object.keys(validityMap).length
+    ) {
+      return false
+    }
+    // Every share must have a validation result of `true`.
```
DarianM, line 144: remove the comment please, there are some unused imports in this file that need cleaning up too

frontend/app/routes/prob-revshare.tsx
```diff
@@ -222,7 +233,9 @@ function Revshare() {
           </div>
         </Card>
         <div className="my-lg md:my-md">
-          <RevShareChart shares={shares} />
+          {revSharePointers && hasValidShares && (
```
DarianM, line 236: isn't  just `hasValidShares` enough for this check? 🤔

DarianM (changes requested):
`handle remove` / onRrmove share in prob-revshare doesn't clean up the `validityMap` when a share is removed,  still contains stale validation results.
to reproduce this, have 3 valid recipients, then invalidate any one of these 3 and then delete the invalid one using the delete button.
we are left with 2 valid recipients, so the link and chart should generate

frontend/app/components/redesign/hooks/useDebounceValidation.ts
```diff
@@ -0,0 +1,81 @@
...
+      error: null
+    }))
+
+    const validatePointer = async (pointer: string) => {
+      let finalState: Pick<ValidationState, 'isValid' | 'error'>
+
+      try {
+        await validateAndConfirmPointer(pointer)
```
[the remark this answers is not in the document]
kjmitchelljr: I can see that, especially since it's happening while user is typing, versus responding to a button click event. Let me play around with this idea and see what I come up with

PR #252, opened 2025-08-20
fix(frontend): fetch configs using interledger.cards WA

DarianM:
### Description

The tools could not retrieve configurations saved using a card wallet because the address was not normalized.


sidvishnoi (approved):
I cannot test as I don't have credentials for this, but looks reasonable and what we used to do earlier.

Issue #174, opened 2025-07-21
[BUG] Banner  |  preview thumbnail


tselit:
### Steps to reproduce

1. Used this link to test: [https://88973536-publisher-tools.webmonetization.workers.dev/tools/banner](https://88973536-publisher-tools.webmonetization.workers.dev/tools/banner)
2. Enter a wallet address that I've never tested with
3. Click on 'Continue'
4. Scroll down to 'Build'
5. Click on 'Content' to expand the section
6. Select all the text in the 'Custom title' textbox and overwrite it with new text
7. Go to the 'Banner message' and overwrite all the text with some new text
8. Click 'Appearance'
9. Edit the colors of the background and the text
10. Change the position of the banner to 'Top'
11. Uncheck the 'Animated' checkbox
12. Click 'Preview'
13. Uncheck the Thumbnail 'Visible' checkbox 
14. Click 'Preview'

### Expected result

The preview shows the banner without a thumbnail when the Thumbnail 'Visible' checkbox is unchecked.

### Actual result

The banner's preview appears the same when the Thumbnail 'Visible' checkbox is checked and when it is unchecked.

### Screenshots or videos

The video is fairly long, you can view the thumbnail behaviour from the 1 minute, 10 second mark: https://www.loom.com/share/85c26054fa9549b0afbdab5e755b35c0?sid=af12ef00-bde7-4cf9-be18-da23fe07a3dc

### Affected component

Banner

### Device type

Desktop

### Browsers

Chrome

### Browser version

Version 138.0.7204.158 (Official Build) (64-bit)

PR #253, opened 2025-08-21
fix(banner): make thumbnail visible checkbox work

sidvishnoi:
Closes https://github.com/interledger/publisher-tools/issues/174

### DIFF
--- a/frontend/app/components/redesign/components/AppearanceBuilder.tsx
+++ b/frontend/app/components/redesign/components/AppearanceBuilder.tsx
@@ -38,6 +38,7 @@ interface BaseToolAppearance {
   buttonColor?: string
   borderRadius?: CornerType
   slideAnimation?: SlideAnimationType
+  thumbnail?: string
 
   onFontNameChange: (fontName: FontFamilyKey) => void
   onFontSizeChange: (fontSize: number) => void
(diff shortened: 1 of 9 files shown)


frontend/app/components/redesign/components/AppearanceBuilder.tsx
```diff
@@ -86,7 +88,9 @@ export const AppearanceBuilder: React.FC<AppearanceBuilderProps> = ({
 }) => {
   const minFontSize = 12
   const maxFontSize = 20
-  const [isThumbnailVisible, setIsThumbnailVisible] = useState(true)
+  const [isThumbnailVisible, setIsThumbnailVisible] = useState(
```
DarianM, line 91: `isThumbnailVisible` remains stale when switching between configurations using tabs, doesn't change state between visible or not visible

sidvishnoi: Good catch!

DarianM:
@sidvishnoi  this initial [commit](https://github.com/interledger/publisher-tools/pull/253/commits/a0724ee0142357d2d62aec2785762248da0a7fd2) solved the rendering issue

PR #256, opened 2025-08-21
fix(frontend): builder accordion button done

DarianM:
### Description

If you click "Done" from the open accordion, it should close and be marked as completed.


sidvishnoi (approved):
LGTM!

There's a lint warning to clear before merging:
```
'ToolsSecondaryButton' is defined but never used. Allowed unused vars must match /^_/u
```

PR #263, opened 2025-08-22
fix: fix rendering issues in builder; fix BuilderAccordion markup

sidvishnoi:
Extracted from https://github.com/interledger/publisher-tools/pull/253

- Move `key` to higher level in `BuilderForm` to correctly re-render on tab change
- Fix markup of `BuilderAccordion`, as `<summary>` isn't allowed to contain `<div>`
- Reduce state in `BuilderAccordion`, making it simpler (no `ref` or `useEffect` needed anymore)
- Maintain open section (content/appearance) when preset tab is switched.

### DIFF
--- a/frontend/app/components/redesign/components/AppearanceBuilder.tsx
+++ b/frontend/app/components/redesign/components/AppearanceBuilder.tsx
@@ -64,11 +64,8 @@ export type ToolAppearance = BannerToolAppearance | WidgetToolAppearance
 interface AppearanceBuilderProps {
   appearance: ToolAppearance
   onRefresh: () => void
-  onDone: () => void
-  isComplete?: boolean
   positionSelector?: React.ReactNode
   colorsSelector?: React.ReactNode
-  activeVersion?: string
 }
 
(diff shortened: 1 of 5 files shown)


frontend/app/components/redesign/components/AppearanceBuilder.tsx
```diff
@@ -114,11 +109,13 @@ export const AppearanceBuilder: React.FC<AppearanceBuilderProps> = ({
...
     <BuilderAccordion
       title="Appearance"
-      isComplete={isComplete}
-      activeVersion={activeVersion}
+      isComplete={uiState.appearanceComplete}
       onToggle={handleToggle}
       onRefresh={onRefresh}
-      onDone={onDone}
+      onDone={() => {
+        uiActions.setAppearanceComplete(true)
```
DarianM, line 116: is it worth opening the other section similar to `handleToggle` if not marked as "completed" ?

sidvishnoi: If not marked as complete, maybe the done button should be disabled?

Issue #189, opened 2025-07-24
[BUG] Banner customizations not applied


Anca2022:
### Steps to reproduce

1. Go to 'https://98bb2f04-publisher-tools.webmonetization.workers.dev/tools/banner'
2. Enter wallet address
3. Click on 'Continue'
4. Scroll down to 'Content' and enter custom content
5. Click on 'Done'
6. Scroll down to 'Appearance' and customise the appearance
7. Click on "Save and generate script"
8. Include the script generated into your HTML file 

### Expected result

A banner that looks like the one from the 'https://98bb2f04-publisher-tools.webmonetization.workers.dev/tools/banner' preview

### Actual result

The banner is not customised; both the content (title + banner message), as well as the appearance do not change 
Screenshot with the configuration: 

<img width="589" height="478" alt="Image" src="https://github.com/user-attachments/assets/3b7f00e5-c403-498d-9174-1f18105c5e81" />

Screenshot with the result: 

<img width="575" height="475" alt="Image" src="https://github.com/user-attachments/assets/265e31f0-eeb2-4e94-b4e0-eb225a23bee2" />

### Affected component

Banner

### Device type

Desktop

### Browsers

Chrome

sidvishnoi:
Most likely relates to https://github.com/interledger/publisher-tools/issues/183.

Issue #183, opened 2025-07-23
Correctly, safely & consistently build, deploy and reference workers


sidvishnoi:
## Context:

Presently, all 3 worker deployments use production endpoints, even in PR previews. For example, if we modify CDN & API, it'll reference production API endpoint instead of the just modified API preview deploy. So, in our preview, if API has some changes, we won't have them available until it's released - defeating purpose of previews.

Also, any malicious user can deploy a worker, and print out our secrets in its HTTP response. There's no proper safety to prevent that. Given we use production secrets all across, it's quite risky.

## To do:

- Set custom github action to deploy all workers as modified
- Use build time env vars to correctly reference dependency URLs (can't do this for secrets of course). For example, When API is modified, deploy CDN too, and use the freshly deployed API endpoint in CDN, so that CDN doesn't use production endpoint.
- Remove cloudflare integration in favor of above
- Automatic deploy previews only on PRs from maintainers. For others, require a manual approval.
- When PR modified, create new deploy versions and provide URLs to deploys.
- When push to `main` branch, deploy new version and set alias to `staging`.
- When push to `release` branch, deploy production version.
- Deploy only modified code and its dependencies.

Issue #135, opened 2025-06-11
fix(ci): Add Build Process to CI/CD Workflow


DarianM:
## Description

Currently, our CI/CD pipeline does not include the build step when a pull request is created. To ensure that code changes do not break the prod, we should update our GitHub Actions workflow to include the build process for all pull requests.

PR #265, opened 2025-08-25
feat: rework deployments, use build-time vars, support staging env

sidvishnoi:
- Closes https://github.com/interledger/publisher-tools/issues/183
- Closes https://github.com/interledger/publisher-tools/issues/189
- Closes https://github.com/interledger/publisher-tools/issues/135

---

- Convert runtime env vars to build time constants
   - now each build/deploy can have a different value of these constants, instead of just the production one.
   - Note: doesn't affect secrets. They're shared and stored within Cloudflare.
   - Keep only two such constants: `API_URL` and `CDN_URL` (earlier script embed URL)
- Remove `FRONTEND_URL` env var from api, and instead pass `redirectUrl`
   - This helps us remove a cyclic dependency in deployments.
   - frontend passes `redirectUrl` to component (even though not directly used) making use of the `Origin` header
   - embed passes `redirectUrl` to component based on the `API_URL` variable.
- Add a `shared/defines` local package, which at build-time replaces these constants
   - Dev remains unaffected due to this helper, without needing any env level changes.
- Given each worker deployment may reference another worker's endpoint, set up a custom CI with GitHub actions to sequentially deploy workers.
   - The `deploy-worker` local action builds and deploys a worker and returns the deployment URL.
      - The worker may not actually deploy (depending on `changed`) input and in that case, may return a staging or production URL.
   - The deploy workflow adds a comment to PRs with deploy links ([example](https://github.com/interledger/publisher-tools/pull/265#issuecomment-3223295252)). For pushes, there's a handy workflow summary added.
   - Deploys only required workers based on files changed.
- For pull requests, a new preview deployment is created (if needed; otherwise staging endpoints are reused).
- For pushes to `main` branch, a new version is deployed and aliased to a staging URL.
- For pushes to `release` branch, new deployment is done to production.

---

To reviewer: First merge https://github.com/interledger/publisher-tools/pull/269 (as it's extracted out of this PR).
Relevant [Slack thread](https://interledgerfoundation.slack.com/archives/C08FYR2GSPP/p1756134118209619).

---

TODO: Once this PR is merged:
   1. Disable git intergration in Cloudflare dashboard
   2. Remove above env vars from each wrangler.toml
      - Ensure production isn't affected by doing this.


shared/defines/index.ts
```diff
@@ -0,0 +1,16 @@
+declare const BUILD_API_URL: string
+declare const BUILD_CDN_URL: string
+
+const DEV_API_URL = 'http://localhost:8787'
+const DEV_CDN_URL = 'http://localhost:5173'
+const _DEV_FRONTEND_URL = 'http://localhost:3000/tools/' // contains trailing slash and pathname
```
kjmitchelljr, line 6: Just curious if this variable is meant to be used anywhere in the code?

api/src/utils/open-payments.ts
```diff
@@ -206,7 +203,7 @@ export class OpenPaymentsService {
       receiveAmount: args.receiveAmount,
       nonce: clientNonce,
       paymentId: paymentId,
-      redirectUrl: this.frontendUrl + 'payment-confirmation'
+      redirectUrl: args.redirectUrl
     })
```
kjmitchelljr, line 207: If I understand correctly, this `redirectUrl` is now coming from the client. Is there any validation that needs to happen on our end or does that take place elsewhere?

shared/defines/index.ts
sidvishnoi: No, it's unused, added for consistency only. So in the future, we know how these things are used.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,155 @@
...
+        uses: ./.github/workflows/deploy-worker
+        id: api
+        with:
+          name: 'api'
+          changed: ${{ steps.changes.outputs.api }}
+          mode: ${{ steps.deploy-mode.outputs.mode }}
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          stagingUrl: 'https://staging-publisher-tools-api.webmonetization.workers.dev'
+          productionUrl: 'https://api.webmonetization.org'
```
raducristianpopa, line 81: Probably worth having the URLs as repository variables? This way if something changes you will only have to update the repository variable.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,155 @@
...
+        run: |
+          mode="preview"
+          if [[ "$EVENT_NAME" == 'push' ]]; then
+            [[ "$GH_REF" == 'refs/heads/main' ]] && mode="staging"
+            [[ "$GH_REF" == 'refs/heads/release' ]] && mode="production"
+          fi
+          echo "mode=${mode}" >> $GITHUB_OUTPUT
+        env:
+          GH_EVENT_NAME: ${{ github.event_name }}
+          GH_REF: ${{ github.ref }}
```
raducristianpopa, line 58: IIRC, inline interpolation can be used for `ref` and `event_name` in the bash script. If this is going to be something consistent - assigning to environment variables, I see no issue here.

components/src/widget/views/confirmation/confirmation.ts
```diff
@@ -267,15 +267,17 @@ export class PaymentConfirmation extends LitElement {
...
     const url = new URL('/tools/payment/grant', apiUrl).href
+    const redirectUrl = new URL('payment-confirmation', frontendUrl).href
 
     const response = await fetch(url, {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json'
       },
       body: JSON.stringify({
+        redirectUrl,
```
raducristianpopa, line 280: Just a thought, not related to CI - maybe it is going to be worth adding an error log for the bad request branch to get more context about the error (I do not know how the API is currently build but maybe more information can help for bug reports latern on).

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,155 @@
...
+      # cdn (embed) needs: API_URL, APP_URL (infer from API_URL to prevent cycle)
+      # frontend (app) needs: CDN_URL, ?API_URL
+      # api needs: nothing (APP_URL is passed from embed)
+      #
+      # build/deploy sequence:
+      # 1. api
+      # 2. cdn (embed)
+      # 3. frontend
+
+      - name: Deploy API
```
raducristianpopa, line 71: Why does the API not have a build command? 🤔

sidvishnoi: It uses the default build command from wrangler (internally esbuild) as it's simpler.

.github/workflows/deploy.yml
sidvishnoi: I wanted to have them available as variables, so used env variables - then I can copy the script out and test it locally.
Perhaps, instead of env vars, I'll see if I can instead use interpolation at the top level of the script - consistency is key.

raducristianpopa (approved):
CI looks good to me.

kjmitchelljr (approved):
Looks good!

PR #275, opened 2025-08-28
fix(ci/deploy): typo in EVENT_NAME when finding deploy mode

sidvishnoi:
Missed in https://github.com/interledger/publisher-tools/pull/265. Meant staging/production deploy mode was never getting set.


sidvishnoi:
I'll merge this without review as it's a small but urgent change.

PR #280, opened 2025-09-02
refactor(frontend): don't nest React component definition

sidvishnoi:
That component will get defined fresh on each render.
Will check why linter didn't raise alarm on that.

Using this PR to check how frontend-only deployment works.


sidvishnoi:
Merging as small change and I want to test CI before bug bash.

Issue #260, opened 2025-08-22
[BUG] RevShare  |  Navigation Dropdown - list in focus on mobile


kjmitchelljr:
### Steps to reproduce

1. Go to Probabilistic Revenue Sharing page
2. Open dev tools
3. Click 'Toggle Device Toolbar'
4. Select any Mobile dimensions
5. Inspect behavior

### Expected result

Only the Tools link should be highlighted, not the entire list

### Actual result

The entire list is being brought into focused when clicking the dropdown

### Screenshots or videos

<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/6d5f5054-a50f-4c6c-a566-08b957b45d88" />

### Affected component

Revenue Share, Others

### Device type

Mobile

### Browsers

Chrome

kjmitchelljr:
@sidvishnoi wondering if I can revert some of the changes here while adhering to a11y improvements made https://github.com/interledger/publisher-tools/commit/caa5748ea83cb97ced206f5c3d68e52c2a1ba060 as the change seemed to have caused the bug that is being seen now?

sidvishnoi:
@kjmitchelljr certainly. Though, we most likely want to retain the focus, but make it not visible on mobile (or some way to show focus only if user Tab-ed into the menu).

PR #281, opened 2025-09-04
fix(Header/NavDropdown): don't show focus style on entire nav

kjmitchelljr:
Closes https://github.com/interledger/publisher-tools/issues/260 while trying to maintain a11y changes from https://github.com/interledger/publisher-tools/pull/232


sidvishnoi:
@kjmitchelljr Yes, they should be the same. And same as shown in Figma (if something wrong in website one, we get it fixed there instead of replicating it here).

sidvishnoi:
@kjmitchelljr I'm ok merging this, and we can match with Figma as a follow up.

kjmitchelljr:
Feel free to review, only main change was taking the classes from the button and applying back to the li element to fix the focus on the entire nav

Issue #85, opened 2025-03-28
Fix: rev share validate url before redirect


DarianM:
## Description

Probabilistic revenue generator api needs to restrict allowed redirect URLs:

- only allow URLs that point to valid payment pointers, an option could be implement a URL allowlist
- reject any non-trusted domains from being used as redirect targets.
- enforce a maximum URL length, prevent large Base64 payloads from being processed.
- sanitize API inputs, ensure that inputs match the expected format and do not allow arbitrary URLs.

edent:
Is there scope to expand the spec to have probabilities in the meta element?

For example, something like:

```html
<link rel="monetization" href="https://wallet.example.com/alice@0.7">
<link rel="monetization" href="https://wallet.alt/bob@0.3">
```

Or

```html
<link rel="monetization" href="https://wallet.example.com/alice?probability=0.7">
<link rel="monetization" href="https://wallet.alt/bob?probability=0.3">
```

None of the [existing element attributes](https://html.spec.whatwg.org/multipage/semantics.html#the-link-element) are a good fit for this use case. So perhaps a new one is needed?


```html
<link rel="monetization" href="https://wallet.example.com/alice" probability="0.7">
<link rel="monetization" href="https://wallet.alt/bob" probability="0.3">
```

PR #296, opened 2025-09-10
refactor: add shared probabilistic-revenue-share package

sidvishnoi:
- Part of https://github.com/interledger/publisher-tools/issues/85
- Extracted from https://github.com/interledger/publisher-tools/pull/282

---

- Add shared package with encode/decode functions and some common revshare types
- Use them in frontend
- Remove unused revshare functions from frontend

---

Follow up:
- @sidvishnoi To use the shared function in API, while fixing the security issues
- @DarianM To use the shared `decode` functionality in Remix backend when importing link tag.

### DIFF
--- a/frontend/app/lib/revshare.ts
+++ b/frontend/app/lib/revshare.ts
@@ -1,4 +1,9 @@
-import { generateShareId } from '@shared/utils'
+import {
+  decode,
+  encode,
+  type PayloadEntry
+} from '@shared/probabilistic-revenue-share'
+
 const BASE_REVSHARE_POINTER = '$webmonetization.org/api/revshare/pay/'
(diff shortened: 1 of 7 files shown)


frontend/app/lib/revshare.ts
```diff
@@ -162,9 +132,7 @@ export function sharesToPaymentPointer(shares: Share[]): string | undefined {
     return
   }
 
-  const pointerList = sharesToPointerList(validShares)
-  const encodedShares = base64url(JSON.stringify(pointerList))
-
+  const encodedShares = encode(validShares)
```
DarianM, line 135: cleanup `sharesToPointerList` function since no longer used and flatMap used with `encode` ?

DarianM (approved):

PR #298, opened 2025-09-10
fix(frontend): tool font size validation

### LINKED ISSUE #288: [Bug] Banner/Widget - Generating a Banner with Text below medium size throws error from the backend
Selecting a small size results in this error. A user should be able to choose any font size offered to them as an option without seeing any errors 
Image

Image

DarianM:
Closes #288 

Different font size ranges for

- widget: min value 16 - max value 20
- banner: min value 20 - max value 24

### DIFF
--- a/shared/types/index.ts
+++ b/shared/types/index.ts
@@ -49,6 +49,11 @@ export interface ElementConfigType {
   widgetTriggerIcon: string
 }
 
+export const FONT_SIZE_RANGES = {
+  widget: { min: 12, max: 20 },
+  banner: { min: 16, max: 24 }
+} as const
+
 export const CORNER_OPTION = {
   Light: 'Light',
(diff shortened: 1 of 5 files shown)


shared/types/index.ts
```diff
@@ -49,6 +49,11 @@ export interface ElementConfigType {
   widgetTriggerIcon: string
 }
 
+export const FONT_SIZE_RANGES = {
+  widget: { min: 12, max: 20 },
+  banner: { min: 16, max: 24 }
+} as const
+
```
sidvishnoi, line 56: I don't think this should belong here. Perhaps in validate files?

sidvishnoi: Also so, keep banner and widget as separate constants.

PR #308, opened 2025-09-12
fix(InputField): better align letter-count position with help text

DarianM:
| Before | After |
|--------|-------|
| image | image

### DIFF
--- a/frontend/app/components/redesign/components/InputField.tsx
+++ b/frontend/app/components/redesign/components/InputField.tsx
@@ -60,12 +60,10 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
         />
         {error && <p className="text-xs text-text-error">{displayError}</p>}
         {(helpText || showCounter) && !error && (
-          <div className="flex justify-between gap-xs">
-            {helpText && (
-              <p className="text-xs text-text-secondary">{helpText}</p>
-            )}
+          <div className="flex justify-between gap-xs text-xs text-text-secondary">
+            {helpText && <p>{helpText}</p>}
             {showCounter && maxLength && (
-              <span className="text-xs leading-xs text-text-secondary">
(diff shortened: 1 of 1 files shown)


frontend/app/components/redesign/components/InputField.tsx
```diff
@@ -43,8 +59,17 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
...
         {error && <p className="text-xs text-text-error">{displayError}</p>}
-        {helpText && !error && (
-          <p className="text-xs text-text-secondary">{helpText}</p>
+        {(helpText || showCounter) && !error && (
+          <div className="flex justify-between gap-xs">
+            {helpText && (
+              <p className="text-xs text-text-secondary">{helpText}</p>
+            )}
+            {showCounter && maxLength && (
+              <span className="text-xs leading-xs text-text-secondary">
```
sidvishnoi, line 68: There's a minor alignment issue due to line height difference. Let's either remove in both or add in both.

```suggestion
              <span className="text-xs text-text-secondary">
```

frontend/app/components/redesign/components/InputField.tsx
```diff
@@ -43,8 +59,17 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
           {...props}
         />
         {error && <p className="text-xs text-text-error">{displayError}</p>}
-        {helpText && !error && (
-          <p className="text-xs text-text-secondary">{helpText}</p>
+        {(helpText || showCounter) && !error && (
+          <div className="flex justify-between gap-xs">
```
sidvishnoi, line 63: If both the text have same styles (`text-xs text-text-secondary`), we can add the styles to parent div.

sidvishnoi (approved):

PR #321, opened 2025-09-16
fix(widget): make widget responsive

### LINKED ISSUE #148: [BUG] Widget does not responsively resize to fit smaller screens
## Description

When the browser window is resized or when the widget is viewed on smaller screens (e.g., mobile or narrow desktop windows), the widget content can overflow outside the viewport. This makes it difficult or impossible for users to interact with the widget properly.

### Expected Behavior:

The widget should be fully responsive and adapt its layout or size to always fit within the screen, regardless of window size.

Steps to Reproduce:

   1. Open a page that contains the widget.
   2.  Resize the browser window to a smaller width.
   3. Observe that part of the widget extends beyond the visible screen.

kjmitchelljr:
Fixes #148 issue


components/src/widget/widget.css
```diff
@@ -460,8 +461,25 @@ a:focus-visible {
...
+    left: 0;
     width: 100%;
+    height: 100%;
+    max-height: 100%;
+    z-index: 9999;
+  }
+
+  .content.open.preview-mode {
+    position: absolute;
+    z-index: 10;
```
sidvishnoi, line 478: LGTM. We can also use some negative margins and extra width/height to expand it fully (as the parent contains padding), but it's a bit hacky.

image

kjmitchelljr: The one thing I'm trying to fix is that gap you see between the `widget-body` and the header/footer of the widget. I found the cause and it's because I'm currently doing @media breakpoints to have it fill the space + there are some flex properties causing this behavior. On the live version this doesn't seem to be an issue.
Screenshot 2025-09-18 at 8 26 20 AM

I'm trying to do the following to fix:
```javascript
@container (max-width: 480px) {
  .content.open.preview-mode {
    position: absolute;
    top: 0;
    left: 0;
    width: 100%;
    height: 100%;
    max-height: 100%;
    border-radius: 0;
    z-index: 10;
  }
}
```
Which would have the widget respond to the `#browser-content` dimensions, but having some trouble as it's not behaving as I would expect it to. @sidvishnoi

sidvishnoi: @kjmitchelljr Make sure you're using `container-type: inline-size;` on the container. 

Might want to avoid using container queries on embed though - it's ok in tools editor, but for embeds on websites, we want better browser support.

components/src/widget/widget.css
```diff
@@ -108,8 +108,9 @@ a:focus-visible {
     0 10px 10px -5px rgba(0, 0, 0, 0.04);
   pointer-events: auto;
   visibility: visible;
+  height: auto;
+  max-height: calc(100vh - 8rem);
```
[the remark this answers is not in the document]
kjmitchelljr: I see. Radu mentioned making it scrollable if the content no longer fits in the widget. Could go back to a fixed height with that add, but open to hear your thoughts @DarianM?

components/src/widget/widget.css
kjmitchelljr: Got it, seems to be behaving as expected now. I lowered the breakpoint to 320px since the default in the editor was less than 480px causing it to always take up the full content area

DarianM: > LGTM. We can also use some negative margins and extra width/height to expand it fully (as the parent contains padding), but it's a bit hacky.
> image

I am not convinced with the  negative margins in preview mode because I can't differentiate the colors and border types if I change the configuration a little, it doesn't look good without some spacing, even if it doesn't reproduce on website 100%.
Same "spacing" is applied in the preview for the banner too

DarianM (approved):

Issue #305, opened 2025-09-12
Fix accessibility of preset tabs in banner/widget


sidvishnoi:
Problem: 
- Anything inside `role=tab` is marked as presentation
- So, a tab cannot contain the edit "button" as well as an input.
- So, screen reader users won't even know that there's a button inside the tab and that the preset name is editable.
- The current JS implementation for making the tab title editable is patchy as we well, and not accessible.

Tasks:
- Remove the edit button and input from `role=tab` and move it into the active `role=tabpanel`.
- Position the edit functionality such that visually it looks like we're editing the tab title.
- Add double click on tabpanel button to trigger same edit functionality, as an affordance for mouse users.
- Simplify edit logic while we're there.

PR #326, opened 2025-09-17
fix(BuilderPresetTabs): use more accessible tabs for presets

sidvishnoi:
Closes https://github.com/interledger/publisher-tools/issues/305

### DIFF
--- a/frontend/app/components/redesign/components/BuilderForm.tsx
+++ b/frontend/app/components/redesign/components/BuilderForm.tsx
@@ -58,7 +58,11 @@ export const BuilderForm: React.FC<BuilderFormProps> = ({
 
   useEffect(() => {
     setTabOptions(() => getLatestTabOptions())
-  }, [toolState.isSubmitting, toolState.currentConfig.walletAddress])
+  }, [
+    toolState.isSubmitting,
+    toolState.currentConfig.walletAddress,
+    toolState.currentConfig.versionName
+  ])
 
   return (
     <div className="flex flex-col mt-4">


frontend/app/components/redesign/components/BuilderForm.tsx
```diff
@@ -39,23 +43,27 @@ export const BuilderForm: React.FC<BuilderFormProps> = ({
...
   }
 
+  const getLatestTabOptions = () => {
+    return toolActions.versionOptions.map(({ stableKey: id }) => ({
+      id,
+      label: toolState.configurations[id].versionName,
+      isDirty: toolState.modifiedVersions.includes(id)
+    }))
+  }
+  const [tabOptions, setTabOptions] = React.useState(getLatestTabOptions)
```
DarianM, line 53: the state should update if we retrieve presets after connecting with a wallet

frontend/app/components/redesign/components/BuilderPresetTabs.tsx
```diff
@@ -0,0 +1,319 @@
...
+          <SVGExclamationCircle className="w-4 h-4 text-text-error" />
+        ) : (
+          <SVGEdit className="w-5 h-5 text-text-buttons-default" />
+        )}
+      </button>
+    </form>
+  )
+}
+
+const validateInput = <T extends string>(
```
DarianM, line 300: can we disable the  save action  buttons somehow if error ? maybe using `uiStore` for this ?

sidvishnoi: If user saves with error, the new name won't be used. We commit the rename only after a valid name is submitted. So, maybe can do without disabling entire form save button.

Issue #302, opened 2025-09-11
Stored configuration Versioning and Data migration


DarianM:
## Description

Currently, users may encounter issues when loading saved configurations that contain properties outside the current validation or when adding new properties to the configurations.
(e.g., font sizes that were valid in previous versions but now fall outside the current min/max constraints).
This means that configurations are not scalable if we want to add new options or modify current ones..
We need versioning system to handle configuration migrations and ensure backward compatibility.

## Proposed Solution

implement a versioning system for tool elements and their configurations, including:

### 1. Configuration Versioning Structure

Each custom element configuration will include a top-level version key:

```json
{
  "banner": {
    "fontName": "Arial",
    "fontSize": 18,
    "backgroundColor": "#ffffff",
    "textColor": "#000000",
    "_version": "1.0.2"
  },
  "widget": {
    "fontName": "Helvetica",
    "fontSize": 14,
    "donateAmount": 5,
    "_version": "2.1.0"
  },
  "button": {
    "text": "Support",
    "fontName": "Georgia",
    "_version": "1.0.2"
  }
}
```

### 2. CDN-Based Versioned Assets
Implement a CDN structure for versioned component delivery:
```
cdn.webmonetization.org/tools/
├── init.js (latest entry point)
├── v1.0.2/
│   ├── button.js
│   ├── widget.js
│   └── banner.js
├── v2.0.1/
│   ├── widget.js
│   └── banner.js
└── v2.1.0/
    └── widget.js
```

### Dynamic import strategy
- function based import
```ruby
import { drawWidget } from 'https://cdn.webmonetization.org/tools/v2.1.0/widget.js';
drawWidget(config, walletAddress);
```
- custom element import
```ruby
const { Widget } = await import('https://cdn.webmonetization.org/tools/v2.1.0/widget.js');
customElements.define('interledger-widget', Widget);

const widget = document.createElement('interledger-widget');
widget.setConfig(config);
document.body.append(widget);
```

### Initialization script logic

The `init.js` script should:
- **Parse configuration** to extract component types and versions
- **Dynamic import** for the correct version component
- **Apply migration** if needed for older configuration
- **Render components** with validated configuration

PR #327, opened 2025-09-18
fix: turn `AWS_PREFIX` to build-time constant

sidvishnoi:
Should help with data migrations per env (https://github.com/interledger/publisher-tools/issues/302). We'll be able to use different `AWS_PREFIX` in staging (migrated data) than production (using prefix from when it was deployed).

### DIFF
--- a/api/wrangler.toml
+++ b/api/wrangler.toml
@@ -6,3 +6,6 @@ compatibility_flags = ["nodejs_compat"]
 [vars]
 FRONTEND_URL = "https://webmonetization.org/tools/"
 AWS_PREFIX = "20250717"
+
+[dev]
+port = 8787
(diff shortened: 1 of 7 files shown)


api/wrangler.toml
```diff
@@ -5,4 +5,7 @@ compatibility_flags = ["nodejs_compat"]
 
 [vars]
 FRONTEND_URL = "https://webmonetization.org/tools/"
-AWS_PREFIX = "20250717"
\ No newline at end of file
+AWS_PREFIX = "20250717"
```
sidvishnoi, line 8: Cleaning these vars will be done post-release, to not affect production versions.

DarianM (approved):

Issue #318, opened 2025-09-15
Implement New Configuration Schema


DarianM:
## Problem statement

Our current configuration system uses a legacy flat data structure that creates technical debt and limits our ability to scale the Publisher Tools app effectively.

This monolithic approach couples all tool configurations together and lacks proper versioning, making it difficult to evolve individual tools independently.

Related to #307 #302 stories

### Current pain points

- **For dev team**
  - tight coupling: all tools share the same configuration interface, making change risky
  - no versioning: new tool styling properties migrations are complex to add, no versioning, difficult testing
- **For product and users**
  - new tools feature are hindered or blocked by legacy configuration format constraints
  - cannot properly implement tool specific customization experiences; interactive tools that share same configuration, bugs such as #319 
  - data insights: difficult to analyze usage patterns per tool due to mixed configuration data
  - limited customization, making it hard to fine-tune individual tools independently

### Proposed Solution: hierarchical configuration system
#307 #302

#### Few immediate benefits
- reduced development risk between interactive tools and faster features development (i.e. Offerwall, Button)
- lay groundwork for advanced Tool style customization and edit profile management
- advanced Tool Analytics is possible more easily 
- improved quality of profile configurations stored in the DB
- features scalability

Issue #307, opened 2025-09-12
Configuration Versioning Structure


DarianM:
## Description

We (@sidvishnoi and @DarianM) discussed this structure over a call, to keep config backward compatible (after this change), structured and easy to migrate. Also included some metadata like `$modifiedAt` timestamp.

Define configurations interface
```ts
interface BaseToolConfig {
  $version: string
  $modifiedAt?: string
}

// The versioned-component knows how to map to actual font size.
type BannerFontSize = '2xs' | 'xs' | 's' | 'base' | 'l' | 'xl' | '2xl'
type WidgetFontSize = '3xs' | BannerFontSize | '3xl'

type HexString = string
type GradientCssString = string

type TextColor = HexString
type Background = HexString | { gradient: GradientCssString }

export default () => (
  <ContentBuilder>
    <ChangeListener bind="title.text">
      <TitleInput
        label="Your title"
        placeholder="some placeholder"
        value={config.title.text}
      />
    </ChangeListener>

    <div className="flex">
      <CheckboxInput
        label="Show description"
        bind:value={config.description.visible}
      />
      <TextAreaInput
        label="Your description"
        placeholder="some placeholder"
        disabled={!config.description.visible}
        bind:value={config.description.text}
      />
    </div>
  </ContentBuilder>
)

interface ButtonConfig extends BaseToolConfig {
  // content
  title: { text: string }
  description: { text: string; visible: boolean }

  // appearance
  fontSize: BannerFontSize
  textColor: TextColor
  background: Background
}

interface WidgetConfig extends BaseToolConfig {
  fontSize: WidgetFontSize
  textColor: TextColor
  background: Background
}

// like we've version1, version2, version3
type PresetIds = 'a' | 'b' | 'c'

type Config = {
  $walletAddress: string
  $walletAddressId?: string
  $modifiedAt?: string
  $presets: {
    [presetId in PresetIds]: { name: string }
  }

  button: {
    [presetId in PresetIds]?: ButtonConfig
  }
  widget: {
    [presetId in PresetIds]?: WidgetConfig
  }
}

export const c: Config = {
  $walletAddress: 'https://interledger.cards/007',
  $walletAddressId: 'https://interledger.cards/007',
  $modifiedAt: '2020-01-01T00:00:00.000Z',
  $presets: {
    a: { name: 'Preset 1' },
    b: { name: 'Default Preset 2' },
    c: { name: 'Default Preset 3' }
  },

  button: {
    a: {
[...]

Issue #88, opened 2025-04-02
Publisher Tools Refactoring & Architecture Overhaul


DarianM:
# Description
As developers working on Publisher Tools, we want a modernized, well-structured codebase with clear separation between frontend and backend, so that the tools are more maintainable, scalable, and easier for new contributors to work with.

Target outcomes:

- Clear separation between frontend and backend with separate domains
- Simplified local development setup without complex Docker configurations
- Modular architecture with shared packages
- Consistent TypeScript usage with well-defined types
- CI/CD pipeline for automated deployments
- Successful migration / rebuilding of Link and RevShare generators from WICG repo 
- Support for Open Payments integration in Widget and Button components
- Web components for custom elements and versioning
- Init script overhaul

### 1. Create comprehensive diagram of the current structure  to guide architecture refactoring efforts.
- [ ] Identify all environment variables currently in use and their purpose
- [ ] Map/Analyze current Docker configuration in `docker-compose.yml` and related files
- [ ] Current deployment processes
   - `wmtools-backend.interledger-test.dev` backend
   - `wmtools.interledger-test.dev` frontend
   - `webmonetization.org/api` for the rev share, through a webmon.org gh actions worker
   - `cdn.webmonetization.org` for the script

### 2. Development Environment Simplification.
- [ ] Replace complex Docker setup with a simpler pnpm-based development environment, have pnpm script for common dev tasks
- [ ] Consolidate environment variables into a single `.env.example` file
- [ ] Implement environment validation at startup, env validation prevents startup with missing critical values
- [ ] Linting and formatting
- [ ] Test setup process with developers who haven't worked on the Tools

### 3. Architecture Overhaul and Frontend/Backend Separation
- [ ] Transitioning deployment to easier solutions (cloudflare)
- [ ] Define API contracts between frontend and backend
- [ ] Backend structure should be  integrated  into Remix, rather than a second backend, current complexity may not require separate backends as currently used
- [ ]  Domain structure:
   - Frontend application: `webmonetization.org.pages.dev/tools/*`
   - API endpoints: `api.webmonetization.org.pages.dev`
   - init script: `cdn.webmonetization.org`
- [ ] Domain structure implemented with clear routing strategy
- [ ] Design shared package structure for common code
- [ ] Have architecture decision records (ADRs) for major decisions

### 4. Shared Package Structure Implementation
- [ ] We should have two packages.jon frontend&backend but released together(the changelog should tell which one we modified)
[...]

PR #337, opened 2025-09-22
feat(api)!: get config per tool; change request structure

sidvishnoi:
Part of https://github.com/interledger/publisher-tools/issues/307, https://github.com/interledger/publisher-tools/issues/318, https://github.com/interledger/publisher-tools/issues/88, https://github.com/interledger/publisher-tools/issues/302.

- Return config only for the given preset and tool, nothing else.
- BREAKING: Instead of `/config/:wa/:version?`, use `/config/:tool?wa={walletAddress}&preset={preset}`.
   - Example: https://pr337-publisher-tools-api.webmonetization.workers.dev/config/banner?wa=https://ilp.interledger-test.dev/sid&preset=version1 | [localhost version](http://localhost:8787/config/banner?wa=https://ilp.interledger-test.dev/sid&preset=version1)
   - Note we don't need to escape wallet address in query param like we had to in path param.
- Make `tool`, `wa` and `preset` required, given the client is our CDN.
- Improved 404 handling (instead of throwing a HTTP 500 from try-catch, manually throw HTTP 404)
- Move route handler to separate file.


api/src/routes/get-config.ts
```diff
@@ -0,0 +1,91 @@
...
+    'param',
+    z.object({
+      tool: z.enum(['widget', 'banner'])
+    })
+  ),
+  zValidator(
+    'query',
+    z.object({
+      wa: z.string().url(),
+      preset: z.enum(['version1', 'version2', 'version3'])
```
DarianM, line 29: would be nice to keep the presets in @shared\types, as only one source of truth;
then will know to modify presets name in one place only. in shared types:
```ts
export const PRESET_IDS = ['version1', 'version2', 'version3'] as const

export type PresetId = (typeof PRESET_IDS)[number]
```

DarianM (approved):

PR #359, opened 2025-09-24
chore(frontend): remove unused assets

DarianM:
Unused svgs, leftover utility functions,  presets


frontend/app/lib/types.ts
```diff
@@ -6,17 +6,24 @@ import type {
 } from '../utils/validate.server.js'
 import type { ElementConfigType } from '@shared/types'
 
-export interface CreateConfigRequest {
-  walletAddress: string
-  tag: string
-  version?: string
-}
+export const modalTypes = [
```
sidvishnoi, line 9: I think we can use this as `type` instead of `const` now. Given when we use it, we don't refer to `modalTypes` `typeof modalTypes` only.

sidvishnoi (approved):

PR #373, opened 2025-10-01
chore(ci/deploy): run on contributors PR

DarianM:
### Context

Maintainers can trigger deployments preview by reviewing with `ok-to-deploy`

`dorny/paths-filter@v3` action does not support the `pull_request_review` event [issue #232](https://github.com/dorny/paths-filter/issues/232)

Work around, changes implement a **manual file change detection** using `git diff` for review-triggered deployments.

Triggered when changes are made to the following file patterns:

### API deployment
- `.github/workflows/deploy.yml` 
- `.github/workflows/deploy-worker/**`
- `api/**` 
- `shared/**`

### CDN deployment
- `.github/workflows/deploy.yml` 
- `.github/workflows/deploy-worker/**`
- `api/**` - API source code (`/cdn` depends on API)
- `shared/**` 
- `components/**`
- `cdn/**`

### Frontend deployment
- `.github/workflows/deploy.yml` 
- `.github/workflows/deploy-worker/**`
- `api/**`
- `shared/**`
- `components/**`
- `cdn/**` - CDN worker (`/frontend` depends on CDN)
- `frontend/**` 

---

**Note:**  deploy.yml matches the pattern ^\.github/workflows/deploy\.yml, all three conditions evaluate to true, causing all modules to deploy on deploy.yml changes.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -81,26 +81,25 @@ jobs:
           # Get list of changed files in the PR
           CHANGED_FILES=$(git diff --name-only $BASE_SHA $HEAD_SHA)
 
-          WORKFLOW_CHANGED=$(echo "$CHANGED_FILES" | grep -qE '^\.github/workflows/(deploy\.yml|deploy-worker/)' && echo true || echo false)
+          SHARED_CHANGED=$(echo "$CHANGED_FILES" | grep -qE '^(\.github/workflows/(deploy\.yml|deploy-worker/)|shared/)' && echo true || echo false)
           API_CHANGED=$(echo "$CHANGED_FILES" | grep -qE '^api/' && echo true || echo false)
-          SHARED_CHANGED=$(echo "$CHANGED_FILES" | grep -qE '^shared/' && echo true || echo false)
           COMPONENTS_CHANGED=$(echo "$CHANGED_FILES" | grep -qE '^components/' && echo true || echo false)
           CDN_CHANGED=$(echo "$CHANGED_FILES" | grep -qE '^cdn/' && echo true || echo false)
           FRONTEND_CHANGED=$(echo "$CHANGED_FILES" | grep -qE '^frontend/' && echo true || echo false)
 
-          if $WORKFLOW_CHANGED || $API_CHANGED || $SHARED_CHANGED; then
+          if $SHARED_CHANGED || $API_CHANGED; then
             echo "api=true" >> $GITHUB_OUTPUT
           else
             echo "api=false" >> $GITHUB_OUTPUT
           fi
 
-          if $WORKFLOW_CHANGED || $API_CHANGED || $SHARED_CHANGED || $COMPONENTS_CHANGED || $CDN_CHANGED; then
+          if $SHARED_CHANGED || $API_CHANGED || $COMPONENTS_CHANGED || $CDN_CHANGED; then
             echo "cdn=true" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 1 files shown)


.github/scripts/check-deploy-permissions.js
sidvishnoi, line 1: We can use TypeScript files directly
https://github.com/interledger/web-monetization-extension/issues/1188

.github/workflows/deploy.yml
```diff
@@ -20,7 +20,29 @@ jobs:
...
+        if: steps.deploy-check.outputs.should-deploy == 'false'
+        run: pnpm -C api run build
+        
+      - name: Build CDN (validation)
+        if: steps.deploy-check.outputs.should-deploy == 'false'
+        run: pnpm -C cdn run build
+        
+      - name: Build Frontend (validation)
+        if: steps.deploy-check.outputs.should-deploy == 'false'
+        run: pnpm -C frontend run build
```
sidvishnoi, line 45: Maybe instead of adding these, we can add a flag to `./.github/workflows/deploy-worker` whether or not it should try to deploy.

.github/scripts/check-deploy-permissions.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {object} params.context - GitHub Actions context
+ */
+module.exports = async ({ core, context }) => {
+
+  // check if author is authorized
+  if (context.eventName === 'pull_request') {
+    const event = /** @type {PullRequest} */ (context.payload);
+    const authorAssociation = event.pull_request.author_association;
+
+    if (!isAllowedAuthor(authorAssociation)) {
```
sidvishnoi, line 20: Can we trigger deployments on manual `pull_request_review` comments from maintainers like we do for extension? Maybe as a follow-up.

package.json
```diff
@@ -33,7 +36,7 @@
   },
   "pnpm": {
     "overrides": {
-      "wrangler": "4.54.0"
+      "wrangler": "4.59.2"
```
sidvishnoi: Avoid this change in this PR please. Made `pnpm-lock.yaml` too big.

DarianM (commented):
ok-to-deploy

.github/workflows/deploy.yml
```diff
@@ -50,6 +70,36 @@ jobs:
...
+          CHANGED_FILES=$(git diff --name-only $BASE_SHA $HEAD_SHA)
+
+          # Check each component (matching the logic from paths-filter)
+          if echo "$CHANGED_FILES" | grep -qE '^(\.github/workflows/(deploy\.yml|deploy-worker/)|api/|shared/)'; then
+            echo "api=true" >> $GITHUB_OUTPUT
+          else
+            echo "api=false" >> $GITHUB_OUTPUT
+          fi
+
+          if echo "$CHANGED_FILES" | grep -qE '^(\.github/workflows/(deploy\.yml|deploy-worker/)|api/|shared/|components/|cdn/)'; then
```
sidvishnoi: Can we keep the grep's simpler like we do with paths-filter? Create a variable for `$SHARED_CHANGED`, `$API_CHANGED`, `$CDN_CHANGED`, and use like:
```
if $SHARED_CHANGED || $API_CHANGED:
  echo "api=true" >> $GITHUB_OUTPUT
else:
  echo "api=false" >> $GITHUB_OUTPUT
```

sidvishnoi (commented):
ok-to-deploy

PR #393, opened 2025-10-07
feat(widget): poll for grant interaction status

### LINKED ISSUE #154: [BUG] Widget's payment won't work in no-opener mode
### Steps to reproduce

1. Add widget script to a website that has set `Cross-Origin-Opener-Policy: same-origin` header
2. Try take a payment from widget

### Expected result

Payment completes when we reach the /tools/payment-confirmation page.

### Actual result

Payment hangs indefinitely as `window.opener` is null here:
https://github.com/interledger/publisher-tools/blob/d9872bb0cadefc37fda1c69f67b007b5e7034278/frontend/app/routes/payment-confirmation.tsx#L22-L24

### Screenshots or videos

_No response_

### Additional context

From MDN:
> Having a Cross-Origin-Opener-Policy header with a value of same-origin prevents setting opener. Since the new window is loaded in a different browsing context, it won't have a reference to the opening window.

### Affected component

Payment Button

### Device type

Desktop

### Browsers

Chrome

### Browser version

_No response_

DarianM:
Closes #154

### DIFF
--- a/components/src/widget/views/interaction/interaction.ts
+++ b/components/src/widget/views/interaction/interaction.ts
@@ -70,7 +70,7 @@ export class PaymentInteraction extends LitElement {
     AbortSignal.timeout(POLLING_TIMEOUT).addEventListener('abort', (ev) => {
       this.#pollingAbortController?.abort(ev)
     })
-    this._startLongPolling({ signal: this.#pollingAbortController.signal })
+    this._startLongPolling(this.#pollingAbortController.signal)
   }
 
   disconnectedCallback() {


api/src/routes/payment.ts
```diff
@@ -91,6 +91,76 @@ app.post(
   }
 )
 
+app.get('/payment/status/:paymentId', async ({ req, json, env }) => {
```
sidvishnoi: Use the `zValidator`.

frontend/app/routes/payment-confirmation.tsx
```diff
@@ -8,21 +15,58 @@ export const meta: MetaFunction = () => {
...
 }
 
+export async function loader({ request, context }: LoaderFunctionArgs) {
+  const { env } = context.cloudflare
+  const url = new URL(request.url)
+  const params: { [key: string]: string } = {}
+
+  url.searchParams.forEach((value, key) => {
+    params[key] = value
+  })
```
sidvishnoi: Can do something like:
```suggestion
  const params = Object.fromEntries([...url.searchParams])
```

frontend/app/routes/payment-confirmation.tsx
```diff
@@ -8,21 +15,58 @@ export const meta: MetaFunction = () => {
...
+
+  try {
+    const existingData = await env.PUBLISHER_TOOLS_KV.get(
+      KV_PAYMENTS_PREFIX + paymentId
+    )
+
+    if (existingData) {
+      return json({ success: true, message: 'Already stored', params })
+    }
+
```
sidvishnoi: Do we care differently anyway if it's already stored?

components/src/widget/views/interaction/interaction.ts
```diff
@@ -108,26 +117,104 @@ export class PaymentInteraction extends LitElement {
...
+      try {
+        const { paymentId } = this.configController.state
+        const { apiUrl } = this.configController.config
+        const url = new URL(`/payment/status/${paymentId}`, apiUrl).href
+
+        const res = await fetch(url, { signal })
+
+        if (res.ok) {
+          const data = (await res.json()) as PaymentStatus
+          switch (true) {
```
sidvishnoi: While I'm personally a fan of `switch (true)`, other people reading the code might get confused. Let's stick to regular if-else.

shared/types/payment.ts
sidvishnoi: Given this is specific to api routes, can we export these types from there and add api as a devDependency where it's needed?

components/src/utils.ts
```diff
@@ -75,3 +80,15 @@ export const getWebMonetizationLinkHref = (userAgent: string): string => {
...
+  params: PaymentStatus
+): params is PaymentStatusSuccess {
+  return 'interact_ref' in params
+}
+
+export function isInteractionRejected(
+  params: PaymentStatus
+): params is PaymentStatusRejected {
+  return 'result' in params && params.result === 'grant_rejected'
+}
```
sidvishnoi: If these are used only in widget/views/interaction.ts, let's keep them there only.

frontend/app/utils/validate.server.ts
```diff
@@ -115,3 +120,23 @@ export const validateForm = async (
...
+  paymentId: z.string().min(1, 'Payment ID is required'),
+  result: z.literal('grant_rejected')
+}) satisfies z.ZodType<PaymentStatusRejected>
+
+export const validatePaymentParams = (params: Record<string, string>) => {
+  return (
+    z.union([
+      PaymentStatusSuccessSchema,
+      PaymentStatusRejectedSchema
+    ]) satisfies z.ZodType<PaymentStatus>
```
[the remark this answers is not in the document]
DarianM: like so? 🤔 don't really follow

api/src/routes/payment.ts
```diff
@@ -91,6 +92,47 @@ app.post(
...
+
+        if (status) {
+          return json({
+            type: 'GRANT_INTERACTION',
+            ...status
+          })
+        }
+
+        await waitWithAbort(POLLING_INTERVAL, signal)
+      }
```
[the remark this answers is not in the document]
sidvishnoi: Don't forget to remove the initial delay if you follow this approach.

api/src/routes/payment.ts
```diff
@@ -91,6 +92,44 @@ app.post(
...
+            type: 'GRANT_INTERACTION',
+            ...status
+          })
+        }
+      }
+
+      throw new Error('AbortError')
+    } catch (error) {
+      if (error instanceof Error && error.message === 'TimeoutError') {
+        throw createHTTPException(504, 'Payment status polling timeout', {})
```
sidvishnoi: Maybe use HTTP 408 here?

PR #394, opened 2025-10-07
refactor(frontend): remove indirection in banner builder & preview

sidvishnoi:
Helps with https://github.com/interledger/publisher-tools/issues/307, https://github.com/interledger/publisher-tools/issues/302.

- Create `BannerBuilder` instead of using `ContentBuilder` and `AppearanceBuilder`
   - Use `BannerConfig` directly instead of converting config specific to `ContentBuilder` and `AppearanceBuilder`
   - Now, stored config maps editor config one-to-one; and we don't need actions to modify config.
- Refactor: Extract `BannerPreview` out of `banner.tsx`
- Turn `BuilderForm` to just `BuilderTabs` - it'll only manage "tabs" (profiles)

Notes to reviewer: most of it just moving code.

### DIFF
--- a/frontend/app/components/banner/BannerBuilder.tsx
+++ b/frontend/app/components/banner/BannerBuilder.tsx
@@ -32,9 +32,9 @@ import {
 } from '~/assets/svg'
 import wmLogo from '~/assets/images/wm_logo_animated.svg?url'
 import { useState } from 'react'
+import { toolState } from '~/stores/toolStore'
 
 interface Props {
-  profile: BannerConfig
   onRefresh: (section: 'content' | 'appearance') => void
 }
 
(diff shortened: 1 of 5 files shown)


kjmitchelljr (commented):
Looks good overall, a lot of moving parts, but clearly more organized.

frontend/app/components/banner/BannerBuilder.tsx
```diff
@@ -0,0 +1,332 @@
...
+          uiActions.setContentComplete(true)
+        }
+      }}
+      onRefresh={() => onRefresh('content')}
+      onDone={() => {
+        uiActions.setContentComplete(true)
+      }}
+      initialIsOpen={uiState.activeSection === 'content'}
+    >
+      <div className="flex flex-col gap-lg">
```
sidvishnoi, line 90: @DarianM Can you check if we do need it (`flex flex-col gap-lg`)? It was in ContentBuilder, but not in ApperanceBuilder.

Issue #218, opened 2025-08-06
a11y: form fields must be associated with labels


sidvishnoi:
The form label is using `<h4>` and there's no link between label and input right now.

<img width="1439" height="406" alt="Image" src="https://github.com/user-attachments/assets/eb5d95ba-2616-49d7-a47f-6b2129614491" />

Should use `<label>` and associate it with its `<input>` instead of using headings. This will also help with expectation that clicking form label should focus input.
Make sure the content of `label` is valid markup.

Similarly, there are places where we need to use `fieldset`+`legend` elements.

sidvishnoi:
Let's keep #452 outside the scope of this.

PR #403, opened 2025-10-09
fix(app): use fieldset and correct label markup for inputs

sidvishnoi:
Part of https://github.com/interledger/publisher-tools/issues/218


kjmitchelljr (approved):
LGTM 👍🏿!

PR #407, opened 2025-10-10
fix(app): state management interaction tools

DarianM:
### Changes
Fix state subscription pattern in banner builder, causing the component to not behave whenever any property in the configuration object changed.

- Read data from snapshots passed down as prop, otherwise use the source proxy state to change state.
- Do not pass `profile` as prop to the `BannerProfile` instead use snapshot in component as it re-rendered the entire route.
- Incorect import and use of `UIStore` provider caused entire app to re-render

### DIFF
--- a/frontend/app/components/banner/BannerPreview.tsx
+++ b/frontend/app/components/banner/BannerPreview.tsx
@@ -22,7 +22,7 @@ export const BannerPreview = ({
   ref
 }: React.PropsWithChildren<Props>) => {
   const [isLoaded, setIsLoaded] = useState(false)
-  const [profile] = useCurrentConfig({ sync: true })
+  const [profile] = useCurrentConfig()
   const bannerContainerRef = useRef<HTMLDivElement>(null)
   const bannerElementRef = useRef<BannerElement | null>(null)
 
--- a/frontend/app/components/widget/WidgetPreview.tsx
+++ b/frontend/app/components/widget/WidgetPreview.tsx
@@ -15,7 +15,7 @@ export const WidgetPreview = ({
   opWallet
 }: React.PropsWithChildren<Props>) => {
   const [isLoaded, setIsLoaded] = useState(false)
-  const [profile] = useCurrentConfig({ sync: true })
+  const [profile] = useCurrentConfig()
   const widgetRef = useRef<WidgetComponent>(null)
 
   useEffect(() => {


frontend/app/components/banner/BannerBuilder.tsx
```diff
@@ -61,8 +59,10 @@ export function BannerBuilder({ onRefresh }: Props) {
 }
 
 function ContentBuilder({ onRefresh }: Props) {
-  const { actions: uiActions, state: uiState } = useUI()
-  const profile = toolState.currentConfig as BannerConfig
+  const uiState = useUIState()
+  const uiActions = useUIActions()
+  const snap = useCurrentConfigSnapshot()
+  const profile = toolState.currentConfig
```
sidvishnoi: Can we add another hook here for this? I would rather we not access `toolState` directly anywhere, and rely on hooks (essentially getters) to maintain abstraction.

frontend/app/hooks/useCurrentConfigSnapshot.ts
sidvishnoi: Maybe export the hook from toolStore only? Then all store related things get exported from same store file.

frontend/app/components/banner/BannerThumbnailSelector.tsx
```diff
@@ -3,16 +3,17 @@ import { Checkbox, Thumbnail } from '@/components'
 import wmLogo from '~/assets/images/wm_logo_animated.svg?url'
 
 interface BannerThumbnailSelectorProps {
-  isVisible: boolean
+  thumbnail: string
   onVisibilityChange: (visible: boolean) => void
```
sidvishnoi: Maybe we can use a `onChange` prop only then?
```ts
value: string;
onChange: (newValue: string) => void
```

frontend/app/components/banner/BannerPreview.tsx
```diff
@@ -6,24 +6,24 @@ import React, {
...
   ref?: React.Ref<BannerHandle>
 }
 
 export const BannerPreview = ({
-  profile,
   cdnUrl,
   ref
 }: React.PropsWithChildren<Props>) => {
   const [isLoaded, setIsLoaded] = useState(false)
+  const snap = useSnapshot(toolState.currentConfig)
```
sidvishnoi: Maybe rename it to `profile` only to avoid too many changes.

sidvishnoi (changes requested):
Let's bring in changes for widget route as well, otherwise it'll be broken if this PR is merged.

frontend/app/components/banner/BannerPreview.tsx
```diff
@@ -6,24 +6,23 @@ import React, {
...
   ref?: React.Ref<BannerHandle>
 }
 
 export const BannerPreview = ({
-  profile,
   cdnUrl,
   ref
 }: React.PropsWithChildren<Props>) => {
   const [isLoaded, setIsLoaded] = useState(false)
+  const [snap] = useCurrentConfigSync()
```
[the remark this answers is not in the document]
sidvishnoi: Yes, just renaming snap to profile locally here. Like profile = snap

frontend/app/stores/toolStore.ts
```diff
@@ -99,6 +100,14 @@ subscribe(toolState, () => {
   updateChangesTracking(toolState.activeVersion)
 })
 
+export function useCurrentConfig(options?: {
+  sync: boolean
+}): [ElementConfigType, ElementConfigType] {
```
sidvishnoi: Is there any place where we don't use `sync: true`?

sidvishnoi (approved):

PR #419, opened 2025-10-16
chore(deps): upgrade React Router to v7

### LINKED ISSUE #129: (refactor) frontend: update remix to react router v7
### Description 

Upgrade to React Router 7, which represents the next major version of Remix with significant architectural improvements

kjmitchelljr:
1. Resolves #129 for updating the front end to React Router v7 from Remix
2. Closes https://github.com/interledger/publisher-tools/pull/427

### DIFF
--- a/frontend/app/routes/api.config.$type.ts
+++ b/frontend/app/routes/api.config.$type.ts
@@ -52,7 +52,7 @@ export async function loader({ request, params, context }: Route.LoaderArgs) {
       // @ts-expect-error TODO
       if (error.name === 'NoSuchKey' || error.message.includes('404')) {
         // no user config exists for this wallet address - return empty response
-        return {}
+        return data({})
       }
       throw error
     }
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -16,7 +16,7 @@
     "@interledger/open-payments": "^7.1.3",
     "@noble/ed25519": "^3.0.0",
     "@paralleldrive/cuid2": "^2.2.2",
-    "@react-router/cloudflare": "^7.0.0",
+    "@react-router/cloudflare": "^7.9.5",
     "@shared/config-storage-service": "workspace:^",
     "@shared/probabilistic-revenue-share": "workspace:^",
     "@shared/utils": "workspace:^",
@@ -36,16 +36,16 @@
     "react-colorful": "^5.6.1",
     "react-dom": "^19.1.1",
     "react-minimal-pie-chart": "^9.1.1",
-    "react-router": "^7.0.0",
+    "react-router": "^7.9.5",
     "sanitize-html": "^2.17.0",
     "tailwindcss": "^3.4.17",
     "valtio": "^2.1.8",
     "zod": "^3.25.76"
   },
   "devDependencies": {
-    "@cloudflare/vite-plugin": "^1.13.13",
+    "@cloudflare/vite-plugin": "^1.14.1",
     "@lit/react": "^1.0.8",
-    "@react-router/dev": "^7.0.0",
+    "@react-router/dev": "^7.9.5",
     "@shared/default-data": "workspace:^",
     "@shared/defines": "workspace:^",
     "@shared/types": "workspace:^",
@@ -61,7 +61,7 @@
     "publisher-tools-api": "workspace:^",
     "react-scan": "^0.4.3",
     "typescript": "5.9.2",
-    "vite": "^7.1.10",
+    "vite": "^7.2.2",
     "vite-tsconfig-paths": "^5.1.4",
     "wrangler": "^4.40.0"
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -165,8 +165,8 @@ importers:
         specifier: ^2.2.2
         version: 2.2.2
       '@react-router/cloudflare':
-        specifier: ^7.0.0
-        version: 7.9.4(@cloudflare/workers-types@4.20251011.0)(react-router@7.9.4(react-dom@19.1.1(react@19.1.1))(react@19.1.1))(typescript@5.9.2)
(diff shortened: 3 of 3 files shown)


frontend/app/routes.ts
kjmitchelljr: > [!NOTE]
> One of the newer files added in upgrade, gave me some issue, but appears to be working fine now. One thing that was recommended was changing `api.config.$type.ts` to `.tsx` file, but not sure if required?

frontend/vite.config.ts
```diff
@@ -1,54 +1,16 @@
...
-        v3_lazyRouteDiscovery: true
-      }
-    }),
-    tsconfigPaths(),
-    devRedirectPlugin()
-  ],
+  plugins: [cloudflare(), reactRouter(), tsconfigPaths()],
   resolve: {
     alias: {
       'crypto': 'crypto-browserify',
```
kjmitchelljr: @DarianM added back in per the comment: https://github.com/interledger/publisher-tools/pull/432#discussion_r2454529305 I'll have to figure this out. It works fine in staging, but on local dev I get the following:
Screenshot 2025-10-23 at 10 55 11 AM
You are correct, it does seem to be an issue with the OP import, specifically 
`import { isGrantValidAndAccepted } from '~/utils/open-payments.server'`

frontend/vite.config.ts
```diff
@@ -1,55 +1,23 @@
...
-      serverModuleFormat: 'esm',
-      future: {
-        v3_fetcherPersist: true,
-        v3_relativeSplatPath: true,
-        v3_throwAbortReason: true,
-        v3_singleFetch: true,
-        v3_lazyRouteDiscovery: true
+    cloudflare({
+      persistState: {
+        path: '../.wrangler/v3'
```
kjmitchelljr: @DarianM I wanted to confirm if this was the correct path. When I run `pnpm preview` I see `.wrangler/state/v3`

Screenshot 2025-10-24 at 1 45 20 PM

DarianM: the correct path for `.wrangler` is the root, that's because the `frontend` should write in the `state/v3` and the `api` reads  data from that same `state/v3`  dir, we need to make sure both point to the same `.wrangler` dir that should be created in root with dev or preview scripts

frontend/package.json
```diff
@@ -4,20 +4,19 @@
...
-    "build": "remix vite:build",
+    "build": "react-router build",
     "predev": "node -e \"require('fs').copyFileSync('../.dev.vars', '.dev.vars');\"",
-    "dev": "remix vite:dev --port 3000",
+    "dev": "react-router dev --port 3000",
     "preview": "pnpm run build && wrangler dev",
     "deploy": "wrangler deploy",
     "prepare": "wrangler types",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "react-router typegen && tsc --noEmit"
```
sidvishnoi: Should `react-router typegen` be part of `prepare` script above?

frontend/app/routes/api.grant.$type.tsx
sidvishnoi:
```suggestion
import type { Route } from './+types/api.grant.$type'

export async function loader({ request, params, context }: Route.LoaderArgs)
```

kjmitchelljr: Got it, QQ if you can help me understand, would we apply this same code `~/routes/api.config.$type.ts` file? Also, wondering whether or not this file should be changed to a `.tsx` or remain `.ts`?

DarianM: let's apply these .react-router types to the other files too, like to `api.config.$type.ts`, since we now have them


Re:
> ts vs tsx: I don't have a preference.

since `api.grant.$type.tsx` is SSR route, it is just server-only code that does not render JSX
; it is a best practice to use `.ts` for server-side route files

frontend/package.json
```diff
@@ -37,15 +36,16 @@
...
     "tailwindcss": "^3.4.17",
     "valtio": "^2.1.8",
     "zod": "^3.25.76"
   },
   "devDependencies": {
-    "publisher-tools-api": "workspace:^",
+    "@cloudflare/vite-plugin": "^1.13.13",
     "@lit/react": "^1.0.8",
-    "@remix-run/dev": "^2.17.1",
+    "@react-router/dev": "^7.0.0",
```
sidvishnoi: Can upgrade to `^7.9.5`.

frontend/package.json
```diff
@@ -4,20 +4,19 @@
...
+    "prepare": "wrangler types && react-router typegen",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
     "@interledger/open-payments": "^7.1.3",
     "@noble/ed25519": "^3.0.0",
     "@paralleldrive/cuid2": "^2.2.2",
-    "@remix-run/cloudflare": "^2.17.1",
-    "@remix-run/react": "^2.17.1",
+    "@react-router/cloudflare": "^7.0.0",
```
sidvishnoi: Can upgrade to `^7.9.5`.

frontend/app/routes/api.config.$type.ts
```diff
@@ -50,17 +47,17 @@ export async function loader({ request, params, context }: LoaderFunctionArgs) {
...
       fileContent = filterDeepProperties(fileContent) as ConfigVersions
 
-      return json(fileContent)
+      return data(fileContent)
     } catch (error) {
       // @ts-expect-error TODO
       if (error.name === 'NoSuchKey' || error.message.includes('404')) {
         // no user config exists for this wallet address - return empty response
-        return json({})
+        return {}
```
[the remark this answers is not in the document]
kjmitchelljr: Great catch, think it was just missed when going through and updating

PR #430, opened 2025-10-22
Test

kjmitchelljr:


frontend/vite.config.ts
```diff
@@ -5,6 +5,7 @@ import tsconfigPaths from 'vite-tsconfig-paths'
 import { APP_BASEPATH } from './app/lib/constants.js'
 import path from 'path'
 
+// Testing something on CF
```
DarianM: hey @kjmitchelljr please use the `pnpm preview` script from the  frontend package if we need to test how will cloudflare behave.
The `preview` is our "staging environment" that closely mimics how the app will behave when deployed to Cloudflare workers.

PR #432, opened 2025-10-22
feat(deps): upgrade to React Router v7 and migrate application code

### LINKED ISSUE #129: (refactor) frontend: update remix to react router v7

kjmitchelljr:
Resolves #129. This pull request migrates the frontend from Remix to React Router v7 for both client and server rendering.

**Framework migration:**

* Replaced all `@remix-run/*` imports and usage with their `react-router` and `@react-router/dev` equivalents in entry files (`entry.client.tsx`, `entry.server.tsx`, `root.tsx`, `worker.ts`), including switching to `HydratedRouter` and `ServerRouter` components for client and server entry points.
* Updated route configuration to use the new `@react-router/dev/routes` API in `routes.ts`.
* Added a new `react-router.config.ts` for React Router's SSR and build configuration.

**Build tooling and scripts:**

* Updated `package.json` scripts to use `react-router` CLI commands for build, dev, and type generation; replaced Remix dependencies with React Router equivalents and updated Vite and related dependencies. 
* Refactored `vite.config.ts` to use `@react-router/dev/vite` and `@cloudflare/vite-plugin` plugins, removing custom Remix and dev proxy plugins, and updated build output directory. 

**TypeScript configuration:**

* Updated `tsconfig.json` to include `.react-router/types` for type generation and set `rootDirs` accordingly. 

**Cloudflare worker integration:**

* Refactored `worker.ts` to dynamically import the React Router server build artifact and update the request handler logic for the new framework.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -84,7 +84,7 @@ importers:
         version: 5.9.2
       wrangler:
         specifier: 4.40.0
-        version: 4.40.0(@cloudflare/workers-types@4.20250507.0)
+        version: 4.40.0(@cloudflare/workers-types@4.20251011.0)
 
   cdn:
     devDependencies:
@@ -123,7 +123,7 @@ importers:
         version: 5.9.2
       wrangler:
         specifier: 4.40.0
-        version: 4.40.0(@cloudflare/workers-types@4.20250507.0)
+        version: 4.40.0(@cloudflare/workers-types@4.20251011.0)
 
   components:
     dependencies:
@@ -161,12 +161,9 @@ importers:
       '@paralleldrive/cuid2':
         specifier: ^2.2.2
         version: 2.2.2
-      '@remix-run/cloudflare':
-        specifier: ^2.17.1
-        version: 2.17.1(@cloudflare/workers-types@4.20250507.0)(typescript@5.9.2)
-      '@remix-run/react':
-        specifier: ^2.17.1
-        version: 2.17.1(react-dom@19.1.1(react@19.1.1))(react@19.1.1)(typescript@5.9.2)
+      '@react-router/cloudflare':
+        specifier: ^7.0.0
+        version: 7.9.4(@cloudflare/workers-types@4.20251011.0)(react-router@7.9.4(react-dom@19.1.1(react@19.1.1))(react@19.1.1))(typescript@5.9.2)
       '@shared/config-storage-service':
         specifier: workspace:^
         version: link:../shared/config-storage-service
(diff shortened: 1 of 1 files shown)


frontend/app/routes.ts
kjmitchelljr: > [!NOTE]
> Newly added due to upgrade to React Router v7, comments may not be necessary

frontend/vite.config.ts
```diff
@@ -1,57 +1,18 @@
...
-      if (req.url === '/') {
-        res.writeHead(302, { Location: APP_BASEPATH + '/' })
-        res.end()
-        return
-      }
-      next()
-    })
-  }
-})
-
```
kjmitchelljr: > [!NOTE]
> **Reason for change:** read it wasn't needed as the `react-router.config.ts` would handle redirect since with `basename: APP_BASEPATH`

frontend/vite.config.ts
```diff
@@ -1,57 +1,18 @@
...
-        v3_lazyRouteDiscovery: true
-      }
-    }),
-    tsconfigPaths(),
-    devRedirectPlugin()
-  ],
+  plugins: [cloudflare(), reactRouter(), tsconfigPaths()],
   resolve: {
     alias: {
-      'crypto': 'crypto-browserify',
```
kjmitchelljr: > [!NOTE]
> **Reason for change:** this was giving me an issue when trying to run locally, in the browser I got the following error -`Error: exports is not defined` and was due to crypto using CommonJS modules. When I removed I was able to run on local dev environment

frontend/vite.config.ts
```diff
@@ -66,6 +27,7 @@ export default defineConfig({
   build: {
     assetsDir: `${APP_BASEPATH.replace(/^\//, '')}/assets`,
     sourcemap: true,
-    target: 'esnext'
+    target: 'esnext',
+    outDir: 'build'
```
kjmitchelljr: This was added because the `cloudflare()` plugin was building the files in `dist` folder instead of `build`

frontend/vite.config.ts
DarianM: We use `nodejs_compat` [flag ](https://developers.cloudflare.com/workers/runtime-apis/nodejs/)in our .toml files to enable nodejs api compatibility for our deployed cloudflare workers, the flag basically gives access to call `crypto` on workers.


There are two different `crypto` implementation out there, Web crypto API and the node.js crypto 
The problem is the OP SDK currently only uses the nodejs runtime crypto: https://github.com/interledger/open-payments-node/issues/4 while the clf worker in production uses the Web crypto API.

This `crypto` incompatibility  causes the OP sdk to not work properly, hence the need to polyfill Node.js crypto with `crypto-browserify`.
I am afraid we still need this polyfill, otherwise production does not work

Issue #287, opened 2025-09-08
[Enhancement] Banner/Widget replace check marks with edited icon


RabebOthmani:
Feedback received: an “edit”  icon should show up instead of the check mark and it should be shown only if there are changes from the saved values

![Image](https://github.com/user-attachments/assets/c294306b-2056-457c-a3df-30d6a774141f)

kjmitchelljr:
Believe this can be closed with change here: https://github.com/interledger/publisher-tools/pull/770 @sidvishnoi?

sidvishnoi:
@kjmitchelljr My understanding is we still want the "edited" markers on accordions (whenever edited, and not just on closed accordions). If not, we likely want to remove the existing checkmark as well (what's their purpose?). Please confirm, @RabebOthmani.

RabebOthmani:
Apologies I missed the notification on this one. @kjmitchelljr where can I check the latest on this one please

kjmitchelljr:
@RabebOthmani I re-opened the PR - but the initial change was here: https://pr756-publisher-tools.webmonetization.workers.dev/tools/banner

PR #435, opened 2025-10-25
refactor(BuilderAccordion): replace SVGGreenVector with SVGEdit

Kathuria:
#287 Feedback received: an “edit” icon should show up instead of the check mark and it should be shown only if there are changes from the saved values

Screenshot 2025-10-25 at 11 19 16 PM


kjmitchelljr:
@Kathuria one issue I'm seeing is that the "Edit" icon is showing even if there aren't any changes made. Ideally, would only want to show the icon if any of the configurations have changed, but if nothing has then leave as is

PR #439, opened 2025-11-03
fix(app): improve input field accessibility and aria-description

kjmitchelljr:
Apart of #218 

This pull request refactors how accessibility descriptions are provided for input fields in the redesign components. 

**Accessibility improvements:**
* Added a new `ariaDescription` prop to the `InputField` component, which generates a visually hidden description and sets the appropriate `aria-describedby` attribute automatically
* **Decision:** Renders the aria description element only when there's no error

**Code simplification in `ShareInput`:**
* Updated all usages of `InputField` in `ShareInput` to use the new `ariaDescription` prop instead of manually rendering `<div className="sr-only">` elements for accessibility descriptions
* Removed redundant accessibility attributes (`aria-required`, `aria-invalid`, `aria-describedby`, and `aria-disabled`) in favor of the new prop
* Cleaned up unnecessary `aria-describedby` attributes from action buttons
* Removed all visually hidden description `<div>` elements that were previously used for accessibility

### DIFF
--- a/frontend/app/components/redesign/components/InputField.tsx
+++ b/frontend/app/components/redesign/components/InputField.tsx
@@ -69,9 +69,8 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
             )}
             id={fieldId}
             name={fieldId}
-            required={required}
-            aria-invalid={!!displayError}
-            aria-describedby={ariaDescriptionId}
+            aria-invalid={!!error}
+            aria-describedby={displayError ? 'input-error' : ariaDescriptionId}
             {...props}
           />
           {ariaDescription && !displayError && (
@@ -82,6 +81,8 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
 
           {displayError && (
             <span
+              id="input-error"
+              role="alert"
               className="absolute right-3 top-full
               -translate-y-1/2
               px-1 text-xs text-text-error bg-white"
--- a/frontend/app/components/redesign/Typography.tsx
+++ b/frontend/app/components/redesign/Typography.tsx
@@ -20,6 +20,7 @@ type TypographyProps = {
   variant: TextStyleType
   as?: React.ElementType
   className?: string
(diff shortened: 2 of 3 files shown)


frontend/app/components/redesign/components/revshare/ShareInput.tsx
```diff
@@ -177,11 +176,8 @@ export const ShareInput = React.memo(
             onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
               onChangeName(e.target.value)
             }
-            aria-describedby={`name-description-${index}`}
+            ariaDescription="Optional name for this recipient."
           />
```
sidvishnoi: Consider using `aria-describedby` instead of inline descriptions, as `*by` ones get translated by browsers.

frontend/app/components/redesign/components/InputField.tsx
```diff
@@ -65,8 +72,17 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
             )}
             id={fieldId}
             name={fieldId}
+            required={required}
+            aria-invalid={!!displayError}
+            aria-describedby={ariaDescriptionId}
+            aria-label={ariaLabel}
```
DarianM: afaik when both `aria-label` and `label` exist, aria-label overrides the visible label for screen readers.
we already have a visible `<label htmlFor={fieldId}` , let's remove the `aria-label` param and only use `label` when the case

frontend/app/components/redesign/components/InputField.tsx
```diff
@@ -65,8 +72,17 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
...
             {...props}
           />
+          {ariaDescription && (
```
DarianM: So, the final decision is that the area description should be displayed at all times, even in case of error?
in combination with `aria-invalid` attr? I am okay with that

DarianM (approved):

PR #449, opened 2025-11-25
chore: add eslint import ordering configuration

DarianM:
## Changes proposed
-  Use `eslint-plugin-import` to enforce consistent import ordering across the project
-  Automatic import sorting with the following order:
   1. node built-in modules
   2. external npm packages (React/React Router prioritized)
   3. Interledger packages (`@interledger/*`)
   4. internal aliases (`@/*`, `~/*`)
   5. relative imports
- Alphabetical sorting within each group (case-insensitive)
- Keep imports compact without blank lines between groups

### Testing
Run `pnpm lint`  from root whenever need to auto-fix import ordering in all files.

### DIFF
--- a/frontend/app/components/redesign/Typography.tsx
+++ b/frontend/app/components/redesign/Typography.tsx
@@ -20,6 +20,7 @@ type TypographyProps = {
   variant: TextStyleType
   as?: React.ElementType
   className?: string
+  htmlFor?: string
   children: React.ReactNode
 } & React.HTMLAttributes<HTMLElement>
 
--- a/frontend/app/components/redesign/components/InputField.tsx
+++ b/frontend/app/components/redesign/components/InputField.tsx
@@ -6,6 +6,7 @@ interface InputFieldProps extends InputHTMLAttributes<HTMLInputElement> {
   id?: string
   error?: string | string[]
   helpText?: string
+  ariaDescription?: string
   showCounter?: boolean
   currentLength?: number
 }
@@ -17,6 +18,7 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
       error,
       id,
       helpText,
+      ariaDescription,
       showCounter = false,
       className = '',
       required,
@@ -28,6 +30,8 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(
   ) => {
     const generatedId = useId()
     const fieldId = id || generatedId
+    const ariaDescriptionId =
+      ariaDescription && !error ? `${fieldId}-aria-desc` : undefined
 
     const getDisplayError = (): string | string[] | undefined => {
       if (required && !props.value) {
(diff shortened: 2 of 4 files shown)


eslint.config.js
```diff
@@ -37,6 +41,52 @@ export default [
...
+            },
+            {
+              pattern: '@/**',
+              group: 'internal'
+            },
+            {
+              pattern: '~/**',
+              group: 'internal'
+            }
+          ],
```
kjmitchelljr:
```suggestion
          ],
          'pathGroupsExcludedImportTypes': ['builtin'],
```

Feel free to test, but found this add would make it to where React rules work and prioritize to top

frontend/app/entry.server.tsx
```diff
@@ -4,13 +4,13 @@
  * For more information, see https://remix.run/file-conventions/entry.server
  */
 
+import { isbot } from 'isbot'
+import { renderToReadableStream } from 'react-dom/server'
 import {
   ServerRouter,
   type AppLoadContext,
   type EntryContext
 } from 'react-router'
```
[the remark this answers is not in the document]
DarianM: hmm. don't think `'pathGroupsExcludedImportTypes': ['builtin']` would re-order like this, but i follow you. we should add that

frontend/react-router.config.ts
```diff
@@ -1,5 +1,5 @@
-import type { Config } from '@react-router/dev/config'
 import { APP_BASEPATH } from './app/lib/constants.js'
+import type { Config } from '@react-router/dev/config'
```
sidvishnoi: React router one should be treated as external and kept on top.

DarianM: it depends, tho external, this import from react router is a type and we keep `type` imports at the bottom

eslint.config.js
```diff
@@ -37,6 +41,63 @@ export default [
...
+              group: 'internal',
+              position: 'after'
+            },
+            {
+              pattern: '~/**',
+              group: 'internal',
+              position: 'after'
+            }
+          ],
+          'pathGroupsExcludedImportTypes': ['builtin'],
```
kjmitchelljr:
```suggestion
          'pathGroupsExcludedImportTypes': ['builtin', 'type'],
```

Not sure if we can have the best of both worlds? Adding to the array moves all type imports to bottom of file, e.g. [OptionSelector.tsx](frontend/app/components/redesign/components/OptionSelector.tsx), but would break some of the import ordering in that file for example it moves a type import to bottom, but it's from `'react'` which we want at top

kjmitchelljr (commented):
No major issues, just some minor things to check out. Probably won't find a perfect solution for some of the little nuances, but overall things look good to me 👍🏿

eslint.config.js
DarianM: I chose to add `type` to `pathGroupsExcludedImportTypes` and have all type imports at the end, without breaking the order of the other imports.

The trade-off is that we cannot apply the same order of path groups to the `type` subgroup, only alphabetical order

PR #452, opened 2025-11-28
fix(app): improve accessibillity for wallet address tooltip

DarianM:
Part of #218 

### Changes

- integrate tooltip with [@floating-ui/react](https://floating-ui.com/docs/react)  for positioning and responsive behavior
   -  `floating-ui` is the upgrade of `popper.js` library
- features
   - positioning: defaults to right placement, flips to top/bottom when space is limited for tooltip's width and automatically shifts to stay withing viewport boundaries
- proper ARIA labels and role

### DIFF
--- a/frontend/app/components/redesign/components/Tooltip.tsx
+++ b/frontend/app/components/redesign/components/Tooltip.tsx
@@ -96,12 +96,18 @@ export function Tooltip({ children, label }: TooltipProps) {
             style={{
               left: arrowX,
               top: arrowY,
-              [{
-                top: 'bottom',
-                bottom: 'top',
-                right: 'left',
-                left: 'right'
-              }[placement.split('-')[0]]!]: `-${ARROW_HEIGHT}px`
+              ...(placement.startsWith('top') && {
+                bottom: `-${ARROW_HEIGHT}px`
+              }),
+              ...(placement.startsWith('bottom') && {
+                top: `-${ARROW_HEIGHT}px`
+              }),
+              ...(placement.startsWith('right') && {
+                left: `-${ARROW_HEIGHT}px`
+              }),
+              ...(placement.startsWith('left') && {
+                right: `-${ARROW_HEIGHT}px`
+              })
             }}
           />
         </div>


sidvishnoi (approved):
Approving as the changes are in right direction, though we need to work on tooltips more properly using some existing libraries.

frontend/app/components/redesign/components/Tooltip.tsx
```diff
@@ -1,42 +1,76 @@
...
+      shift({ padding: 8 }),
+      size({
+        apply({ availableWidth, elements }) {
+          const maxWidth = Math.min(availableWidth, 450)
+          Object.assign(elements.floating.style, {
+            maxWidth: `${maxWidth}px`,
+            width: 'auto'
+          })
+        },
+        padding: 8
```
sidvishnoi: Can we have this `8` as a constant and explain why this particular value?

frontend/app/components/redesign/components/Tooltip.tsx
```diff
@@ -1,42 +1,76 @@
...
         className="rounded-full hover:bg-gray-100 focus:outline-none focus:ring-1 focus:ring-primary-focus"
       >
         <SVGTooltip className="w-6 h-6" />
       </button>
 
       {open && (
         <div
-          id={tooltipId}
+          ref={refs.setFloating}
+          id="tooltip"
```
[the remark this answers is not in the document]
DarianM: I don't see the case where multiple tooltips could be open at same time.
We could use the react context to globally manage tooltip id state , but floating-ui doesn't have a setting for that.

frontend/app/components/redesign/components/Tooltip.tsx
```diff
@@ -1,26 +1,76 @@
...
+    <>
+      <button
+        ref={refs.setReference}
+        type="button"
+        aria-label="More information"
+        aria-describedby={open ? 'tooltip' : undefined}
+        onMouseEnter={() => setOpen(true)}
+        onMouseLeave={() => setOpen(false)}
+        onFocus={() => setOpen(true)}
+        onBlur={() => setOpen(false)}
```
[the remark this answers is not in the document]
DarianM: Yes, but we are using the light ui-floating version, so we have to do it ourselves with the `useEffect`

frontend/app/components/redesign/components/Tooltip.tsx
```diff
@@ -1,26 +1,109 @@
...
+  useEffect(() => {
+    const handleEscape = (e: KeyboardEvent) => {
+      if (e.key === 'Escape' && open) {
+        setOpen(false)
+      }
+    }
+
+    document.addEventListener('keydown', handleEscape)
+    return () => document.removeEventListener('keydown', handleEscape)
+  }, [open])
```
sidvishnoi: If we remove the `open` dependency, and call `setOpen(false)` without checking if open, this `useEffect` will not need to register/deregister events on every open/close change?

DarianM: The event listener will only be registered once when the component mounts this way. Improvement works

frontend/app/components/redesign/components/Tooltip.tsx
```diff
@@ -1,26 +1,111 @@
...
+            className="absolute w-4 h-4 bg-interface-tooltip rotate-45"
+            style={{
+              left: arrowX,
+              top: arrowY,
+              [{
+                top: 'bottom',
+                bottom: 'top',
+                right: 'left',
+                left: 'right'
+              }[placement.split('-')[0]]!]: `-${ARROW_HEIGHT}px`
```
sidvishnoi: Took me time to figure out lol. Do you think it'll be easier to reason if we go like:
```js
...(placement.startsWith('top-') && {bottom: `-${ARROW_HEIGHT}px`}),
...(placement.startsWith('bottom-') && {top: `-${ARROW_HEIGHT}px`}),
// ...
```

DarianM: Don't mind if I do :) good in my book

PR #470, opened 2025-12-04
feat(api): add wallet address fetching endpoint

### LINKED ISSUE #467: fix(frontend): move wallet address validation from client to API to simplify CSP requirements
## Problem
Currently, the payment `widget.ts` component directly fetches wallet address information from various wallet provider domains using `getWalletAddress` util. This requires users to add every possible wallet provider domain to their `connect-src` CSP directive, which is:

impractical - users must know and whitelist all potential wallet provider domains
prone to changes- new wallet providers require updating CSP configuration
user-unfriendly - creates unnecessary friction for implementation

### Proposed Solution
Move wallet address validation to a new API endpoint. The widget will call our API, which then fetches from the wallet provider server-side.
New API Route: GET /wallet

```json
// request (as query param for GET maybe?)
{
  "walletAddress": "https://wallet-provider.com/alice"
}

// response
{
  "id": "https://wallet-provider.com/alice",
  "assetCode": "USD",
  "assetScale": 2,
  "authServer": "https://auth.wallet-provider.com",
  "resourceServer": "https://wallet-provider.com",
  "publicName": "Alice's Wallet"
}

// error response
{
  "error": {
    "message": "This wallet address does not exist",
    "code": "WALLET_NOT_FOUND"
  }
}
```

part of #146

kjmitchelljr:
Closes #467 

This pull request introduces a new `/wallet` API endpoint for validating wallet addresses

### DIFF
--- a/frontend/app/components/redesign/components/BuilderPresetTabs.tsx
+++ b/frontend/app/components/redesign/components/BuilderPresetTabs.tsx
@@ -4,7 +4,7 @@ import { SVGEdit, SVGExclamationCircle } from '~/assets/svg'
 
 type TabOption<T extends string> = { id: T; label: string; isDirty: boolean }
 interface Props<T extends string> {
-  options: TabOption<T>[]
+  options: readonly TabOption<T>[]
   children: React.ReactNode
   selectedId: T
   idPrefix: string
@@ -197,7 +197,7 @@ function TabActionTrigger({ onClick }: { onClick: () => void }) {
 }
 
 interface TabNameEditorProps<T extends string> {
-  options: TabOption<T>[]
+  options: readonly TabOption<T>[]
   tabId: T
   tabIdx: number
   onSubmit: (label: string) => void
@@ -312,7 +312,7 @@ function TabNameEditor<T extends string>({
 
 const validateInput = <T extends string>(
   value: string,
-  options: TabOption<T>[],
+  options: readonly TabOption<T>[],
   currentTabId: T
 ): string | '' => {
   const val = value.trim()
--- a/frontend/app/routes.ts
+++ b/frontend/app/routes.ts
@@ -5,6 +5,7 @@ export default [
   index('routes/_index.tsx'),
   // Tool routes
   route('banner/', 'routes/banner.tsx'),
+  route('banner-two/', 'routes/banner-two.tsx'),
   route('link-tag/', 'routes/link-tag.tsx'),
   route('payment-confirmation/', 'routes/payment-confirmation.tsx'),
   route('prob-revshare/', 'routes/prob-revshare.tsx'),
--- a/frontend/app/routes/banner-two.tsx
+++ b/frontend/app/routes/banner-two.tsx
@@ -1,0 +1,393 @@
+import { useEffect, useState, useRef } from 'react'
+import {
+  useLoaderData,
+  useNavigate,
+  data,
+  type LoaderFunctionArgs,
+  type MetaFunction
+} from 'react-router'
+import { useSnapshot } from 'valtio'
+import { SVGSpinner } from '@/assets'
+import {
+  HeadingCore,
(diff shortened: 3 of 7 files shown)


components/src/widget/widget.ts
```diff
@@ -59,19 +72,11 @@ export class PaymentWidget extends LitElement {
...
     }
 
-    let walletAddressInfo: WalletAddress
     try {
-      const walletAddressUrl = checkHrefFormat(
-        toWalletAddressUrl(walletAddress)
-      )
-
-      walletAddressInfo = await getWalletAddress(walletAddressUrl)
+      const walletAddressInfo = await this.validateWalletAddress(walletAddress)
```
DarianM:
```suggestion
const response = await fetch(
  `${API_URL}/wallet?walletAddress=${encodeURIComponent(walletAddress)}`
)

const walletAddressInfo = await response.json()
if (!response.ok) {
  throw new Error(
    walletAddressInfo.error.message
  )
}

this.configController.updateState({
  walletAddress: walletAddressInfo as WalletAddress
})
```

by using, for example, `https://ilp.interledger-test.dev/kjmitchelljr` as a wallet address for an address that does not exist, the widget should display the message “This wallet address does not exist.” returned by the api.
Also, I think it's more cleaner if called directly without creating `validateWalletAddress`

api/src/routes/wallet.ts
```diff
@@ -0,0 +1,41 @@
...
+  getWalletAddress,
+  normalizeWalletAddress,
+  toWalletAddressUrl,
+  WalletAddressFormatError
+} from '@shared/utils'
+import { app } from '../app.js'
+import { createHTTPException } from '../utils/utils.js'
+
+const walletAddressSchema = z.object({
+  walletAddress: z.string().min(1, 'Wallet address is required')
```
sidvishnoi: We can require this to be a URL, as it's called internally by us.

api/src/routes/wallet.ts
```diff
@@ -0,0 +1,41 @@
...
+
+app.get(
+  '/wallet',
+  zValidator('query', walletAddressSchema),
+  async ({ req, json }) => {
+    const { walletAddress } = req.valid('query')
+
+    try {
+      const walletAddressUrl = checkHrefFormat(
+        toWalletAddressUrl(walletAddress)
```
sidvishnoi: I don't think we need `toWalletAddressUrl` (or even `checkHrefFormat`) here as input is a URL. This API is called internally only with particular inputs that we control, and it failing with bad input with fail at `getWalletAddress` itself.

components/src/widget/widget.ts
```diff
@@ -59,20 +54,28 @@ export class PaymentWidget extends LitElement {
...
+      url.searchParams.set('walletAddress', walletAddressUrl)
+
+      const response = await fetch(url)
+
+      if (!response.ok) {
+        const errorData = (await response.json()) as { message: string }
+        throw new Error(errorData.message)
+      }
+
+      const walletAddressInfo: WalletAddress = await response.json()
```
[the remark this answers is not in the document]
kjmitchelljr: I went back and forth on that, I can make the change

components/src/widget/widget.ts
```diff
@@ -59,20 +54,26 @@ export class PaymentWidget extends LitElement {
...
 
-      walletAddressInfo = await getWalletAddress(walletAddressUrl)
+      const url = new URL('/wallet', apiUrl)
+      url.searchParams.set('walletAddress', walletAddressUrl)
+
+      const response = await fetch(url)
+      const data = await response.json()
+
+      if (!response.ok) {
+        throw new Error((data as { message?: string }).message)
```
DarianM: We need to get the message from the `error` object so that the widget can display it. Something like:

```suggestion
        throw new Error(data.error?.message)
```
otherwise, the error message from the API is not displayed

PR #484, opened 2025-12-11
refactor(app): integrate profiles into testbed banner builder

DarianM:
Part of #307 

Notes:
- switch Profile Tabs and fetch profiles into the banner store
- temporary `convertFrom` to store s3 bucket configurations in profile stores;
   - can be later used to do the migrations, probably

### DIFF
--- a/frontend/app/components/banner/BannerPreview.tsx
+++ b/frontend/app/components/banner/BannerPreview.tsx
@@ -7,7 +7,10 @@ import React, {
 } from 'react'
 import type { BannerConfig, Banner as BannerElement } from '@tools/components'
 import { useBannerProfile } from '~/stores/banner'
-import { toolState, useCurrentConfig } from '~/stores/toolStore'
+import {
+  toolState,
+  useCurrentConfig as useCurrentConfigLegacy
+} from '~/stores/toolStore'
 
 export interface BannerHandle {
   triggerPreview: () => void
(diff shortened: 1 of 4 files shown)


frontend/app/components/banner/BannerPreview.tsx
```diff
@@ -6,7 +6,8 @@ import React, {
   useState
 } from 'react'
 import type { BannerConfig, Banner as BannerElement } from '@tools/components'
-import { useCurrentConfig } from '~/stores/toolStore'
+import { useBannerProfile } from '~/stores/banner'
+import { toolState, useCurrentConfig } from '~/stores/toolStore'
```
sidvishnoi:
```suggestion
import { toolState, useCurrentConfig as useCurrentConfigLegacy } from '~/stores/toolStore'
```

frontend/app/components/banner/BannerPreview.tsx
```diff
@@ -17,12 +18,20 @@ interface Props {
   ref?: React.Ref<BannerHandle>
 }
 
+function snapshot(options?: { sync: boolean }) {
+  if (toolState.currentToolType === 'banner-two') {
+    return useBannerProfile(options)
+  }
+
+  return useCurrentConfig(options)
```
sidvishnoi:
```suggestion
function useCurrentConfig(options?: { sync: boolean }) {
  if (toolState.currentToolType === 'banner-two') {
    return useBannerProfile(options)
  }

  return useCurrentConfigLegacy(options)
```

PR #486, opened 2025-12-13
refactor(app): persist testbed store to storage 

DarianM:
Part of #307 

- Persist banner store on refresh

### DIFF
--- a/frontend/app/stores/banner.tsx
+++ b/frontend/app/stores/banner.tsx
@@ -48,7 +48,6 @@ export function useBannerProfile(options?: {
 }
 
 export const banner = createBannerStore()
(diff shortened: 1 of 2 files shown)


frontend/app/stores/banner.tsx
```diff
@@ -45,6 +48,7 @@ export function useBannerProfile(options?: {
 }
 
 export const banner = createBannerStore()
+PROFILE_IDS.forEach((id) => subscribeProfileToStorage(id))
```
sidvishnoi: This statement will be executed as soon as this file is imported, including in tests. I'd suggest moving it to route file only.

PR #491, opened 2025-12-16
refactor(app): persist baseline store to storage

DarianM:
Part of #307 

- persist state differs from its last saved snapshot, indicating unsaved changes (i.e. dirty profile)


frontend/package.json
```diff
@@ -32,6 +32,7 @@
     "httpbis-digest-headers": "^1.0.0",
     "isbot": "^5.1.32",
     "lit": "^3.3.1",
+    "lodash": "^4.17.21",
```
sidvishnoi: Let's not add lodash, as we can get carried away with using it too much when it's here. Maybe https://www.npmjs.com/package/fast-equals would be better alternative - should check impact on bundle size first.

frontend/app/stores/banner/snapshots.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function isProfileDirty(id: ProfileId) {
+  const profile = snapshots.get(id)
+  if (!profile) {
+    return false
+  }
+
+  const snap = snapshot(banner.profiles[id])
+  return !isEqual(snap, profile)
```
sidvishnoi: If we make use of `$modifiedAt`, we can avoid a relatively more expensive `isEqual` all the time.

PR #504, opened 2026-01-08
refactor(app): save individual profile in banner store

DarianM:
Part of #307 
### Proposed changes
- Add route for individual `banner-two` profile save
- `useSaveProfile` hook to use at save cta
- `Banner` zod schema and validations

### DIFF
--- a/frontend/app/stores/banner-store.ts
+++ b/frontend/app/stores/banner-store.ts
@@ -95,12 +95,8 @@ export const actions = {
   },
   async saveProfile(): Promise<SaveResult> {
     const profile = snapshot(banner.profile)
-    return await saveToolProfile(
-      toolState.walletAddress,
-      TOOL_BANNER,
-      profile,
-      toolState.activeTab
-    )
+    const { walletAddress, activeTab } = toolState
+    return await saveToolProfile(walletAddress, TOOL_BANNER, profile, activeTab)
   },
   commitProfile() {
     const profile = snapshot(banner.profile)


frontend/app/routes/api.profile.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  if (!walletAddress) {
+    return data(
+      { error: 'walletAddress query param required' },
+      { status: 400 }
+    )
+  }
+  if (!profileId || !PROFILE_IDS.includes(profileId)) {
+    return data({ error: 'Invalid profileId' }, { status: 400 })
+  }
```
sidvishnoi: Use `zod` based validation here?

frontend/app/routes/api.profile.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (!parsed.success) {
+      const errors = z.treeifyError(parsed.error)
+      return data({ errors }, { status: 400 })
+    }
+
+    const walletAddressData = await getWalletAddress(walletAddress)
+
+    const session = await getSession(request.headers.get('Cookie'))
+    const validForWallet = session.get('validForWallet')
+    session.set('wallet-address', walletAddressData)
```
sidvishnoi: Some attention needed here. If I send this call manually from devtools console, with a different wallet address (than cookie), can I save config for a wallet address that I don't own?

sidvishnoi: Maybe the `.set` should be after we know valid session?

frontend/app/hooks/useSaveProfile.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const [openDialog, closeDialog] = useDialog()
+
+  const save = useCallback(
+    async (action: 'save-success' | 'script'): Promise<void> => {
+      toolState.lastSaveAction = action
+
+      try {
+        const response = await actions.saveProfile()
+
+        if (!response.success && response.grantRequired) {
```
sidvishnoi: Do we want to handle `!response.success` in other cases as well? Or will it be true all other ways?

frontend/app/routes/api.profile.ts
```diff
@@ -0,0 +1,122 @@
...
+      )
+    }
+
+    session.set('wallet-address', walletAddressData)
+
+    const walletAddressId = normalizeWalletAddress(walletAddressData)
+    const storage = new ConfigStorageService({ ...env, AWS_PREFIX })
+    let config: Configuration = {
+      $walletAddress: walletAddress,
+      $walletAddressId: walletAddressId,
```
sidvishnoi: I don't follow. Why not use `walletAddressData.id`?

frontend/app/routes/api.profile.ts
```diff
@@ -0,0 +1,122 @@
...
+      $modifiedAt: new Date().toISOString()
```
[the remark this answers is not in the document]
DarianM: not anymore, please check again when possible

frontend/app/routes/api.profile.$tool.ts
```diff
@@ -0,0 +1,133 @@
...
+import { commitSession, getSession } from '~/utils/session.server.js'
+import { walletSchema } from '~/utils/validate.server'
+import {
+  BannerProfileSchema,
+  WidgetProfileSchema
+} from '~/utils/validate.shared'
+
+const ApiSaveProfileSchema = z.object({
+  ...walletSchema.shape,
+  profile: z.union([BannerProfileSchema, WidgetProfileSchema]),
```
sidvishnoi: How about something like:
```ts
profile: { tool: 'banner', data: BannerProfileSchema } | { tool: 'widget', data: WidgetProfileSchema }
```

We can keep `tool` within body (no need to use a param) and validate data based on tool type.

Presently, what if a malicious user (or even accidentally us) sends `WidgetProfileSchema` when using `param = banner`? Our validation will allow it - it's only checking if either widget or tool, not whether widget tool + widget config.

frontend/app/routes/api.profile.ts
```diff
@@ -0,0 +1,159 @@
...
+    } catch (e) {
+      const err = e as Error
+      if (err.name !== 'NoSuchKey' && !err.message.includes('404')) {
+        throw e
+      }
+
+      config = {
+        $walletAddress: walletAddress,
+        $walletAddressId: walletAddressId,
+        $createdAt: new Date().toISOString(),
```
sidvishnoi: Nit:
```ts
const now = new Date().toISOString()
// then use it
```

frontend/app/stores/banner-store.ts
```diff
@@ -88,6 +93,41 @@ export const actions = {
...
+
+    if (!response.ok) {
+      throw new ApiError(
+        data.error?.message || 'Failed to save profile',
+        data.error?.cause?.errors
+      )
+    }
+
+    return data
+  },
```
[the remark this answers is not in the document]
sidvishnoi: I'd say not require toolState in that function either. Just pass it all the params (wallet address, profile id) similarly.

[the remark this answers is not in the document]
sidvishnoi: I'd say not require toolState in that function either. Just pass it all the params (wallet address, profile id) similarly.

sidvishnoi (approved):

PR #531, opened 2026-01-27
refactor!: allow interactive tools to save in the new format

### LINKED ISSUE #511: refactor: Enable API to save data in new format - Phase 3
## Description
Update storage layer to persist data in the new `Configuration` format instead of legacy `ConfigVersions` format.

### Acceptance Criteria:
  - modify storage service to save Configuration format type

DarianM:
Closes #511

### DIFF
--- a/shared/defines/index.ts
+++ b/shared/defines/index.ts
@@ -5,7 +5,7 @@ declare const BUILD_AWS_PREFIX: string
 const DEV_API_URL = 'http://localhost:8787'
 const DEV_CDN_URL = 'http://localhost:5173'
 const DEV_FRONTEND_URL = 'http://localhost:3000/tools/' // contains trailing slash and pathname.
-const DEV_AWS_PREFIX = '20260126-dev'
+const DEV_AWS_PREFIX = '20260305-dev'
 
 export const APP_URL = {
   production: 'https://webmonetization.org',


api/src/routes/get-profile.ts
```diff
@@ -46,9 +47,21 @@ app.get(
...
-      const profile = convertToProfile(legacyProfile, tool)
+      let profile: ToolProfile<typeof tool> | null = null
+      try {
+        const config = await storage.getJson<Configuration>(walletAddress)
+        profile = config[tool]?.[profileId] ?? null
+      } catch (e) {
+        const err = e as Error
+        if (err.name !== 'NoSuchKey' && !err.message.includes('404')) {
+          throw e
+        }
```
sidvishnoi: No longer using `ConfigStorageServiceError` here?

sidvishnoi (changes requested):
I saved a config from `main` branch, then switch to this PR. In frontend when adding my wallet address to banner tool, I received 404 (Configuration not found), and the default config was shown in editor.

frontend/app/routes/api.profile.ts
```diff
@@ -148,6 +149,10 @@ export async function action({ request, context }: ActionFunctionArgs) {
           $modifiedAt: now,
         },
       },
+    })
+
+    if (hasLegacy) {
+      await storage.delete(walletAddressId, true)
     }
```
sidvishnoi: So, let's add some transaction/rollback feature here, in case either fails, we stay in a consistent state:
- If new one is saved and then old one fails to delete, we'll have trouble at migration time. User will feel config not saved (throw on delete), but actually it got saved.
- If old one is deleted (first) then new one fails to save, user has lost config unless they immediately save again successfully.

shared/config-storage-service/index.ts
```diff
@@ -24,9 +26,12 @@ export class ConfigStorageService {
     this.client = ConfigStorageService.instance
   }
 
-  async getJson<T>(walletAddress: string): Promise<T> {
+  async getJson<T>(walletAddress: string, useLegacy = false): Promise<T> {
     const key = walletAddressToKey(walletAddress)
-    const url = new URL(`${this.prefix}/${key}`, this.endpoint)
+    const url = new URL(
+      `${useLegacy ? this.LEGACY_AWS_PREFIX : this.prefix}/${key}`,
```
sidvishnoi: Nit: `const prefix = ...`.

PR #545, opened 2026-02-03
fix(ci/deploy): ensure pushes to main/release trigger deploys

DarianM:


.github/scripts/check-deploy-permissions.ts
```diff
@@ -40,22 +40,19 @@ export default async function checkDeployPermissions({ core, context, github }:
...
-      owner: context.repo.owner,
-      repo: context.repo.repo,
-      username: context.actor
-    });
-
-    const isAuthorized = ['admin', 'maintain', 'write'].includes(permission.permission);
-
-    if (!isAuthorized) {
-      await skipDeployment(core, 'Not authorized to trigger deployments.');
+    const branch = process.env.BRANCH_NAME;
```
sidvishnoi: Why can check target branch with `event.base_ref`? Check what TypeScript types suggest for the push event.

PR #549, opened 2026-02-06
test(frontend): add validation tests for api profile route

DarianM:
Part of #511

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -43,3 +43,14 @@ jobs:
         uses: ./.github/workflows/setup
       - name: Check format
         run: pnpm format:check
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


frontend/app/routes/api.profile.test.ts
```diff
@@ -0,0 +1,409 @@
...
+      request: mockRequest,
+      context: mockContext,
+    } as ActionFunctionArgs)
+
+    expect(response.init?.status).toBe(400)
+    expect(response.data.error?.message).toContain('widgetTitleText')
+    expect(response.data.error?.cause?.errors).toEqual({
+      reason: SANITIZATION_ERROR_CAUSE,
+      field: 'widgetTitleText',
+    })
```
sidvishnoi: Don't expose these details for malicious users who can check which fields are sanitized, which aren't, and which type of error it is.
Return a generic bad request with "invalid payload".

PR #565, opened 2026-02-14
refactor(widget): use new profile format

DarianM:
Part of #511 

#### Context
Part of the ongoing effort to deprecate `ElementConfigType` in favor of typed tool profiles

- Have `WidgetProfile` in the new format
- Use the new format throughout 
- Save widget in the old format

### DIFF
--- a/api/src/routes/get-profile.ts
+++ b/api/src/routes/get-profile.ts
@@ -15,7 +15,6 @@ import type {
   ElementConfigType,
   Tool,
   ToolProfile,
-  WidgetConfig,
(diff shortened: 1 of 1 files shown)


sidvishnoi (changes requested):
I added the widget from preview deploy to a site, and got this error:

TypeError: can't access property 'theme', t is undefined

Looks like API (from CDN embed) returned response in old format (I see `widgetTitleText` in JSON response) but component expected new format (forgot conversion somewhere?)

Rest looks good 👍

PR #568, opened 2026-02-25
fix(offerwall): keep the appearance builder open

DarianM:
## Changes

- keep Offerwall appearance section open all the time
- rename `BuilderPresetTabs` -> `BuilderProfileTabs`
- remove unused BuilderTabs component
- fix color picker rendering behind Done button

### DIFF
--- a/api/src/routes/get-profile.ts
+++ b/api/src/routes/get-profile.ts
@@ -13,7 +13,6 @@ import type {
   BaseToolProfile,
   ConfigVersions,
   ElementConfigType,
-  OfferwallProfile,
   Tool,
   ToolProfile,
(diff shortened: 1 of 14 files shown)


sidvishnoi (changes requested):
The changes here break a11y. The corresponding `tabpanel` should be there even when not visible.

DarianM:
> The changes here break a11y. The corresponding `tabpanel` should be there even when not visible.

yep. reverted the change entirely

DarianM:
> Seems broken without it @DarianM
> Screen.Recording.2026-02-26.at.15.28.41.mov

yes, i gave up in fixing it now. too much focus

PR #569, opened 2026-02-25
fix(frontend): ensure scroll to wallet address input scrolls/focuses

DarianM:
### Changes 

Extract the scroll hook and use `requestAnimationFrame` api

### DIFF
--- a/frontend/app/hooks/useScrollToWalletAddress.ts
+++ b/frontend/app/hooks/useScrollToWalletAddress.ts
@@ -1,7 +1,9 @@
 import { useCallback, useRef } from 'react'
+import { useUIActions } from '~/stores/uiStore'
 
 export function useScrollToWalletAddress() {
   const walletAddressRef = useRef<HTMLDivElement>(null)
+  const uiActions = useUIActions()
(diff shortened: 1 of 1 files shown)


frontend/app/hooks/useScrollToWalletAddress.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function useScrollToWalletAddress() {
+  const walletAddressRef = useRef<HTMLDivElement>(null)
+
+  const scrollToWalletAddress = useCallback(() => {
+    if (!walletAddressRef.current) {
+      return
+    }
+
+    requestAnimationFrame(() => {
```
sidvishnoi: Why `requestAnimationFrame` is needed?

DarianM: when you trigger the hook on button click, react needs a brief moment to update the error state with what's shown on screen. scrollIntoView was running before that update finished.
The WalletAddress component state was not settled yet, so the scroll landed in the wrong spot sometimes

Issue #482, opened 2025-12-11
[improvement] Implement internationalization (i18n) for gradual content localization


DarianM:
## Context
Currently, all text content in the publisher tools is hard coded directly within components. This makes it:
- scattered text all over places
- difficult to maintain and update copy
- impossible to support multiple languages
- hard to review all user-facing text in one place

Followup of #446 

#### Expectations

- [ ] set up i18n infra using `react-i18next` or `remix-i18next`
- [ ] establish translation key naming conventions
- [ ] migrate components to gradually use translations
- [ ] create locale files with existing text

PR #573, opened 2026-02-27
feat(frontend): i18n boilerplate

DarianM:
Part of #482


kjmitchelljr (approved):
Small nits here and there, but overall the PR looks good

frontend/app/components/redesign/components/ToolsWalletAddress.tsx
```diff
@@ -24,6 +25,7 @@ export const ToolsWalletAddress = ({
   walletActions,
   toolName,
 }: Props) => {
+  const { t } = useTranslation()
```
sidvishnoi: Why not this?
```suggestion
  const t = useTranslation()
```

frontend/app/i18n/useTranslation.ts
```diff
@@ -0,0 +1,42 @@
+import { useI18n, type Translations } from './context'
+
+/**
+ * Helper recursively builds all valid key paths from a nested translation object,
+ * using `__` as the namespace separator so key names can freely contain `_`.
+ * e.g. { a: { b_c: 'x' } } → 'a__b_c'
+ */
+type UnderscorePath<T, Acc extends string = ''> = {
```
sidvishnoi: Nit: Why not  dot notation, like `t('toolsWalletAddress.status.connect')`?

frontend/app/i18n/useTranslation.ts
```diff
@@ -0,0 +1,42 @@
...
+  [K in keyof T & string]: T[K] extends string
+    ? `${Acc}${K}`
+    : T[K] extends Record<string, unknown>
+      ? UnderscorePath<T[K], `${Acc}${K}__`>
+      : never
+}[keyof T & string]
+
+type TranslationKey = UnderscorePath<Translations>
```
sidvishnoi: Think this. If someone wants to edit the locales file, how will they find the key best? Like I want to edit above key (in either _ or . notations), I'll have to manually find first `toolsWalletAddress`, then `status`, then `connect`.
If I could copy-paste entire key and Ctrl+F for it, that's easier for me to edit.

sidvishnoi (commented):
Added a suggestion to consider non-nested objects..

frontend/app/i18n/useTranslation.ts
DarianM: because if any translation key name itself contains a `.` . translations would have problems when using dot.

sidvishnoi: Yes, that explains `__`. What I meant, why `_` separator over `.` (i.e. can have any separator, even `/`).
 But use one format, then key can't have that character.

DarianM: switched to a more flat json

frontend/app/i18n/locales/en.json
```diff
@@ -0,0 +1,16 @@
+{
+  "toolsWalletAddress": {
+    "heading.message": "Wallet address",
+    "tooltip.ariaLabel": "Why connect your wallet?",
+    "tooltip.message": "Connecting your wallet allows us to save your custom profiles, link them to you as the original author, and verify ownership for future updates.",
+    "tooltip.message2": "It also enables Web Monetization on your behalf by embedding your wallet address into your web page automatically.",
+    "errors.fieldRequired": "This field is required",
```
sidvishnoi, line 14: Consider adding a `general`/`default` namespace as well. Then we won't have to add messages like this in each namespace. Even better if keys can fallback to the general namespace.

frontend/app/i18n/useTranslation.ts
```diff
@@ -0,0 +1,21 @@
...
+function interpolate(str: string, params: Params): string {
+  return str.replace(/\{\{(\w+)\}\}/g, (_, key) => params[key] ?? `{{${key}}}`)
+}
+
+export function useTranslation<NS extends Namespace>(ns: NS) {
+  const translations = useI18n()
+
+  function t(key: NamespaceKey<NS>, params?: Params): string {
+    const value = (translations[ns][key] as string) ?? key
+    return params ? interpolate(value, params) : value
```
sidvishnoi, line 28: Need to be super careful here. While `ns`, and `key` are in our control (make sure they are, using TypeScript - like don't allow passing anything other than from our locales file to these functions (`ns` already taken care of)), the value in `params` can be more risky.
 
While React will escape most strings (to reduce XSS risks), we need to be careful to not be using `dangerouslySetInnerHTML` with this, or using in attributes, specially when user-generated params can be involved.
```tsx
<a href={t("{{url}}", { url: "javascript:alert('XSS')" })}>Click Me</a>
```

Also, what if params (if key can be user-generated as well) contains keys like `constructor`? We've prototype injection risk.

frontend/app/components/redesign/components/ToolsWalletAddress.tsx
```diff
@@ -95,26 +99,27 @@ export const ToolsWalletAddress = ({
...
+        message: t('status.connect'),
         type: 'info',
       }
     }
     if (!snap.hasRemoteConfigs) {
       return {
-        message: `There are no custom edits for the ${toolName} correlated to this wallet address but you can start customizing when you want.`,
+        message: t('status.noSavedProfiles', {
+          toolName,
+        }),
```
sidvishnoi: Nit: avoid extra new lines when not making code cleaner to read.
```suggestion
        message: t('status.noSavedProfiles', { toolName }),
```

frontend/app/i18n/useTranslation.ts
DarianM: so in this PR `t('status.connect')` compiles. `t('status.doesNotExist')` is a compile error. TypeScript derives the valid values directly from en.json i.e. if a key isn't in the file, it can't be passed. (`key` is also taken care of). Tested for compile error if wrong key.

the only thing genuinely outside our control is the values inside param, those can come from user input at runtime, which is why added a  `Object.hasOwn` + `String()` guards to not return the built-in `Object` constructor functions like when using {{constructor}}.
i.e. Object.hasOwn(params, key)  should check only for it's own properties.

`t()` returns a plain string and react escapes it safely in JSX text yes., but it can't protect against dangerouslySetInnerHTML usage. maybe add a `t` description with something like "never pass to dangerouslySetInnerHTML or unsanitized href/src" ?

frontend/app/i18n/locales/en.json
DarianM: added a general namespace.
the fallback is:  `namespace translation` key > `general` key > or even plain `key` in json

sidvishnoi (approved):

PR #579, opened 2026-03-02
fix(api): return default profile on 404 instead of error

DarianM:
## Description 

New publishers with no saved profile were receiving a 404 response,
causing the embed to fail on load.

### changes

 `GET /profile/:tool` no longer throws a 404 when no saved profile exists
  for a wallet address — it now falls back to the default profile for the
  requested tool  without requiring a prior save.

### DIFF
--- a/api/src/routes/get-profile.ts
+++ b/api/src/routes/get-profile.ts
@@ -4,7 +4,9 @@ import { zValidator } from '@hono/zod-validator'
 import {
   ConfigStorageService,
   ConfigStorageServiceError,
+  isConfigStorageNotFoundError,
 } from '@shared/config-storage-service'
+import { getDefaultProfile } from '@shared/default-data'
 import { AWS_PREFIX } from '@shared/defines'
 import {
   numberToBannerFontSize,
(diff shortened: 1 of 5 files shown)


api/src/routes/get-profile.ts
```diff
@@ -48,12 +49,11 @@ app.get(
...
-          const msg = 'No saved profile found for given wallet address'
-          throw createHTTPException(404, msg, {
-            message: 'Not found', // can include the S3 key here perhaps
-            code: '404',
-          })
+        if (error.name === 'NoSuchKey' || error.message.includes('404')) {
+          console.warn(
+            `[404] No profile found for ${walletAddress}. Returning default ${tool} profile.`,
+          )
+          return json<ToolProfile<typeof tool>>(getDefaultProfile(tool))
```
sidvishnoi: Maybe we can still set `statusCode: 404` when returning response? Then we can judge 404's from network request insights from Cloudflare, instead of checking logs.

sidvishnoi (approved):

PR #588, opened 2026-03-04
fix(offerwall): align container max-width with Figma spec

DarianM:
See https://github.com/interledger/publisher-tools/pull/583#pullrequestreview-3882433424


frontend/app/routes/offerwall.tsx
```diff
@@ -143,116 +143,111 @@ export default function Offerwall() {
 
           <Divider className="!my-3xl" />
 
-          <div className="flex flex-col min-h-[756px]">
-            <div className="flex flex-col xl:flex-row xl:items-start gap-lg">
```
DarianM: all this indentation because merged two `div` into one

PR #604, opened 2026-03-09
fix(frontend/builder): keep customotization sections open when switching between profile tabs

### LINKED ISSUE #596: Offerwall: Fix switching between open tabs should not close profile sections
Note to Assignee: keep the section always open. see if possible to remove chevron when having only one section in builder

DarianM:
Closes #596


kjmitchelljr (approved):
Changes look good 👍🏿

PR #607, opened 2026-03-10
fix(frontend/builder): prevent refresh button on click to collapse the accordion

DarianM:
#### Context

- prevent refresh button click to bubble up to `summary` on close the accordion
- hide collapse arrow for the offerwall builder that cannot be collapsed 

regression after #605

### DIFF
--- a/frontend/app/components/redesign/components/BuilderAccordion.tsx
+++ b/frontend/app/components/redesign/components/BuilderAccordion.tsx
@@ -59,15 +59,20 @@ export const BuilderAccordion: React.FC<Props> = ({
             <GhostButton
               icon="refresh"
               iconPosition="left"
-              onClick={onRefresh}
+              onClick={(e) => {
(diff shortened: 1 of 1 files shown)


frontend/app/components/redesign/components/BuilderAccordion.tsx
```diff
@@ -59,15 +59,20 @@ export const BuilderAccordion: React.FC<Props> = ({
...
+              }}
               aria-label={`Reset ${title.toLowerCase()} to default`}
             >
               Back to default
             </GhostButton>
           )}
-          <SVGArrowCollapse
-            className={cx('w-12 h-12 p-3.5', !isOpen && 'rotate-180')}
-          />
+          {onDone && (
```
sidvishnoi: Looks like this check isn't enough for Offerwall. I still see the collapse icon:

Image

PR #641, opened 2026-03-26
chore(migration/001): convert legacy configuration to the new format

### LINKED ISSUE #512: refactor: Create utility for migrating stored data formats - Phase 4
## Description
Build migration utility tools to convert existing stored `ConfigVersions` type data to the new `Configuration` format in the database/storage.

 ### Acceptance Criteria
  - create migration script for existing data
  - good practice: add rollback capability for failed migrations
  - good practice: create validation to ensure data integrity post migration

DarianM:
### Summary
**one-time CLI to migrate legacy S3 config data to the new format**

1. Adds a `migrations` CLI that reads wallet configs from the legacy S3 prefix, converts them to the new shape, writes them to the new prefix.
  - wallets already present in the new prefix are cleaned up (legacy key deleted) and counted as skipped.

3. Adds migrations/s3/ as a standalone S3 client package wrapping client-s3, shared by the CLI.

4. Includes a dry-run mode (--dry-run) that simulates all S3 writes/deletes without touching real data, and writes its own separate log file.

5. Each run produces a timestamped log file under migrations recording successful, skipped, and failed wallets.

6. Migration packages are excluded from the root pnpm workspace — they live under their own migrations/pnpm-workspace.yaml and are installed separately (cd migrations && pnpm install), so CI is not affected


migrations/001-config-s3-structure/index.ts
```diff
@@ -0,0 +1,131 @@
...
+      'prefix': `${this.LEGACY_AWS_PREFIX}/`,
+    })
+
+    const response = await this.client.fetch(url)
+    if (!response.ok) {
+      throw new Error(`Failed to list S3 objects: ${response.status}`)
+    }
+
+    // we use aws4fetch raw signed HTTP, so we need to parse the XML response manually to extract the keys
+    const result = xmlParser.parse(await response.text()).ListBucketResult
```
sidvishnoi: There's no list endpoint that returns JSON?

migrations/s3/index.ts
```diff
@@ -0,0 +1,148 @@
...
+function keyToWalletAddress(key: string): string {
+  return `https://${key.replace(/^[^/]+\//, '').replace(/\.json$/, '')}`
+}
+
+export class S3MigrationClient {
+  private readonly client: S3Client
+  private readonly bucket: string
+
+  constructor(config: S3Config) {
+    this.bucket = config.bucket
```
DarianM, line 32: The reason I didn't add this AWS module globally in the root was because the migration only uses it for now.
Swapping the aws library now also requires `bucket` and `region`. we can hardcode or add as dev vars

migrations/001-config-s3-structure/index.ts
DarianM: not with `aws4fetch` . does not provide a list as JSON array abstraction . aws4fetch is thin and have to work directly with the raw s3 HTTP api which returns XML.

migrations/s3/index.ts
DarianM: if the structure overall is good, we can extract in smaller PRs

pnpm-workspace.yaml
```diff
@@ -5,6 +5,7 @@ packages:
   - 'components'
   - 'shared/*'
   - 'localenv/*'
+  - 'migrations/*'
```
sidvishnoi: These and related dependencies will be installed in CI, which won't really be used. Any way we can ignore them in CI, and only install when needed locally?

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,263 @@
...
+  return new S3MigrationClient({
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID!,
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY!,
+    bucket: process.env.AWS_S3_BUCKET!,
+    region: process.env.AWS_S3_REGION,
+    endpoint: process.env.AWS_S3_ENDPOINT,
+  })
+}
+
+export async function dryRun(
```
sidvishnoi: Having a separate function/login as `dryRun` won't really be a dry-run. Please add only one function, and instead pass it a mock-s3 so it acts like dry-run (for non-recoverable actions), while keeping a single flow.

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    const legacyData = await s3.getJson<ConfigVersions>(walletAddress)
+    if (!legacyData) {
+      console.log('! No legacy data found - skipping')
+      return false
+    }
+    const newData = convertToConfiguration(legacyData, walletAddress)
+    await s3.putJson(walletAddress, newData)
+    await s3.deleteFromLegacy(walletAddress)
```
sidvishnoi: What if either of these fails? Can we have something like a transaction here?

sidvishnoi: Can we [batch delete](https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-multiple-objects.html) in end? We should write a local log file to write which files (and/or a cursor) have already been processed, then delete them in end when we're sure.

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,263 @@
...
+      console.log('\n ✓ Batch migration complete')
+    }
+  } catch (error) {
+    console.error('\n x Migration failed:', error)
+    process.exit(1)
+  }
+}
+
+// only run main CLI if this file executed directly
+if (import.meta.url === pathToFileURL(process.argv[1]).href) {
```
sidvishnoi: `import.meta.main`?

pnpm-workspace.yaml
DarianM: sure. introduced different workspace for migrations, but will have to cd into `/migrations`, manually install and run tests.

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: yes, since  node v24.2.0,

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: we'll  manually delete everything after we make sure everything went smoothly. manually or in batch.

DarianM: removed delete.
we'll  manually delete everything after we make sure everything went smoothly. manually or in batch.

migrations/s3/index.ts
```diff
@@ -0,0 +1,148 @@
...
+  GetObjectCommand,
+  HeadObjectCommand,
+  ListObjectsV2Command,
+  PutObjectCommand,
+  S3Client,
+  S3ServiceException,
+} from '@aws-sdk/client-s3'
+
+export const LEGACY_PREFIX = '20250717-dev'
+export const NEW_PREFIX = '20260305-dev'
```
sidvishnoi: These should probably be passed from where they're used. Otherwise, we'll be updating this package whenever we need to migrate, while this data is none of its business (it should just care about a "prefix" or even just key).

migrations/s3/index.ts
```diff
@@ -0,0 +1,148 @@
...
+      }),
+    })
+  }
+
+  private buildKey(prefix: string, walletAddress: string): string {
+    return `${prefix}/${walletAddressToKey(walletAddress)}`
+  }
+
+  async getJson<T>(walletAddress: string): Promise<T | null> {
+    const key = this.buildKey(LEGACY_PREFIX, walletAddress)
```
sidvishnoi: I think class class should simply just get `key: string`. We use `walletAddress` in other place, but that's not "S3" (it's "config generic storage service"), this is S3 util in particular.

migrations/s3/index.ts
```diff
@@ -0,0 +1,148 @@
...
+      new PutObjectCommand({
+        Bucket: this.bucket,
+        Key: key,
+        Body: JSON.stringify(data),
+        ContentType: 'application/json',
+      }),
+    )
+  }
+
+  async listLegacyWallets(): Promise<string[]> {
```
sidvishnoi:
```suggestion
  async listByPrefix(prefix: string): Promise<string[]> {
```

migrations/pnpm-workspace.yaml
sidvishnoi, line 1: I'm not following what this does.
I think the goal was to ensure that `migrations` dependencies don't get installed when we run `pnpm i` from project root (like in CI(), as they're only to be installed when we want to run a migrations.

pnpm-lock.yaml
sidvishnoi, line 1: What caused changes in this file if we added a pnpm-workspace in `migrations`?

migrations/001-config-s3-structure/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "s3"
+  ],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@migration/s3": "workspace:^",
+    "@shared/utils": "workspace:^"
+  },
+  "devDependencies": {
+    "@shared/types": "workspace:^",
```
sidvishnoi, line 22: Will it safely get dependencies from other workspace?

migrations/001-config-s3-structure/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "version": "1.0.0",
+  "description": "Migration CLI for converting legacy ConfigVersions format to new Configuration format",
+  "type": "module",
+  "main": "./migrate-cli.ts",
+  "scripts": {
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "test:coverage": "vitest run --coverage",
+    "typecheck": "tsc --noEmit",
+    "migrate": "tsx --env-file=../../.dev.vars migrate-cli.ts"
```
sidvishnoi: Might not need `tsx`, as `node` CLI typically supports running .ts files directly (if they don't have some "unsafe" TypeScript).

migrations/001-config-s3-structure/migration-log.ts
```diff
@@ -0,0 +1,82 @@
...
+const LOG_DIR = join(import.meta.dirname, '..', 'logs')
+
+function runTimestamp(): string {
+  const iso = new Date().toISOString()
+  const date = iso.slice(0, 10).replace(/-/g, '.')
+  const time = iso.slice(11, 19).replace(/:/g, '')
+  return `${date}.${time}`
+}
+
+// example log filename: migration.2026.04.05.143022.json
```
sidvishnoi: `.gitignore` these files?

migrations/001-config-s3-structure/migration-log.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    if (this.failed.length > 0) {
+      console.log('\nErrors:')
+      this.failed.forEach(({ wallet, error }) => {
+        console.log(`  ${wallet}: ${error}`)
+      })
+    }
+  }
+
+  save(): void {
```
sidvishnoi, line 69: Make sure to call this in case of crash (where used) as well (haven't seen the call site yet).

migrations/001-config-s3-structure/__tests__/migration.test.ts
```diff
@@ -0,0 +1,346 @@
...
+    listLegacyWallets: vi.fn(),
+  })),
+  LEGACY_PREFIX: '20250717-dev',
+  NEW_PREFIX: '20260305-dev',
+}))
+
+const WALLET = '$wallet.example.com'
+
+/** Full legacy profile — all banner + widget fields populated */
+const mockLegacyData: ConfigVersions = {
```
sidvishnoi: Please convert these to functions that return this data. There's a risk if some function modifies a property of these objects, other tests will see modified data and break. Returning a fresh object from a function can prevent those bugs.

migrations/pnpm-workspace.yaml
DarianM: We hadn’t included migrations in the root pnpm workspace, so creating a separate workspace for it made sense.

migrations/001-config-s3-structure/package.json
DarianM: Yes, because `../shared/*` is listed in migrations/pnpm-workspace.yaml

pnpm-lock.yaml
DarianM: forgot to cleanup

migrations/001-config-s3-structure/package.json
DarianM: right. node cli is enough 👍

migrations/s3/index.ts
DarianM: 💯

migrations/s3/index.ts
DarianM: this a batter name. renamed.

migrations/s3/index.ts
DarianM: not 100% follow on this one 🤔

DarianM: S3 only cares about the key now 👍

migrations/001-config-s3-structure/migration-log.ts
DarianM: the root .gitignore already handles it, which matches any logs/ directory anywhere in the repo tree

migrations/001-config-s3-structure/migration-log.ts
DarianM: added a finally block that guarantees save runs whatever, at call site

migrations/001-config-s3-structure/migrate-cli.ts
sidvishnoi, line 1: Let's rename this to `index.ts`

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
...
+    .replace('$', '')
+    .replace('https://', '')
+  return `${prefix}/${normalized}.json`
+}
+
+function keyToWalletAddress(key: string): string {
+  return `https://${key.replace(/^[^/]+\//, '').replace(/\.json$/, '')}`
+}
+
+export interface MigrationClient {
```
sidvishnoi: Should this be exported from s3? like an interface that S3 implements.

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+interface CliOptions {
+  wallet?: string
+  batch: boolean
+  dryRun: boolean
+}
+
+export function parseArgs(): CliOptions {
```
sidvishnoi: Use a lightweight helper package like `sade`, instead of writing own, which can be brittle.

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
...
+    getJson: (key) => client.getJson(key),
+    existsAt: (key) => client.existsAt(key),
+    listByPrefix: (prefix) => client.listByPrefix(prefix),
+    putJson: () => Promise.resolve(),
+    deleteAt: (key) => {
+      console.log(`[DRY RUN] would deleteAt: ${key}`)
+      return Promise.resolve()
+    },
+  }
+}
```
sidvishnoi: Just in case, you can override only the ones you need:
```ts
class DryRunClient extends S3MigrationClient {
  constructor(args) {
    super(args);
  }

  putJson: () => Promise.resolve(),
  deleteAt: (key) => {
    console.log(`[DRY RUN] would deleteAt: ${key}`)
    return Promise.resolve()
  },
}
```

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
...
+
+export async function migrateSingle(
+  s3: MigrationClient,
+  walletAddress: string,
+): Promise<boolean> {
+  const key = walletAddressToKey(NEW_PREFIX, walletAddress)
+  const legacyKey = walletAddressToKey(LEGACY_PREFIX, walletAddress)
+  try {
+    if (await s3.existsAt(key)) {
+      await s3.deleteAt(legacyKey)
```
sidvishnoi: I thought we decided not to delete any files as of yet?

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
...
+    if (!legacyData) {
+      console.log('! No legacy data found - skipping')
+      return false
+    }
+    const newData = convertToConfiguration(legacyData, walletAddress)
+    await s3.putJson(key, newData)
+
+    return true
+  } catch (error) {
+    console.error('x Migration failed:', (error as Error).message)
```
sidvishnoi: Include wallet address in this msg maybe, so we know where exactly it failed.

migrations/001-config-s3-structure/migrate-cli.ts
sidvishnoi: And what if it's already deleted and we call it again? Does it throw or ends silently?

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
...
+    return
+  }
+
+  const log = new MigrationLog(isDryRun)
+  console.log(`\nStarting batch migration for ${wallets.length} wallets...`)
+
+  try {
+    for (const wallet of wallets) {
+      try {
+        const migrated = await migrateSingle(s3, wallet)
```
sidvishnoi: If we've already migrated, we should check the log file and skip running for that wallet address again. That's one reason we want log files - they act like cache for us.

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
...
+  s3: MigrationClient,
+  isDryRun: boolean,
+): Promise<void> {
+  const wallets = (await s3.listByPrefix(LEGACY_PREFIX)).map(keyToWalletAddress)
+  if (wallets.length === 0) {
+    console.log('x No wallets found in legacy prefix')
+    return
+  }
+
+  const log = new MigrationLog(isDryRun)
```
sidvishnoi: Consider passing `log` as an argument to this function; then we can call `migrateBatch().finally(() => log.save())`, instead of nested try-catch blocks.

migrations/pnpm-workspace.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+  - '001-config-s3-structure'
+  - 's3'
```
sidvishnoi, line 3:
```suggestion
  - './*'
```

sidvishnoi: Or do you want to install only for particular migrations the current way? That we'll update the workspace to include new migration and remove old one as already completed?

migrations/001-config-s3-structure/migrate-cli.ts
```diff
@@ -0,0 +1,236 @@
+import { S3MigrationClient } from '@migration/s3'
+import type { ConfigVersions } from '@shared/types'
+import { convertToConfiguration } from '@shared/utils/profile-converter'
+import { MigrationLog } from './migration-log'
+
+const LEGACY_PREFIX = '20250717-dev'
+const NEW_PREFIX = '20260305-dev'
```
kjmitchelljr: Still going through to understand changes, but should these be environment variables or stored in a different way or is this something only going to be ran once and no further change would be required?

migrations/pnpm-workspace.yaml
DarianM: Yes. The current approach explicitly lists specific directories, means we have to manually add new migrations and remove completed ones from the workspace file.
Include just new ones, remove old ones, update this file.

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: I like `sade` but does not automatically check for typos in flags or throw an 'unknown flag' error. added custom small check

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: yeah, that's the right place.

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: yes, I'd keep it as-is and not complicate this time. I think the current factory approach is a good enough fit

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: seemed safe to delete if the wallet was already present in new prefix, means the wallet is already migrated.
you don't want anything to deal with that wallet anyway, to not override stuff.


> And what if it's already deleted and we call it again? Does it throw or ends silently?

but yeah, forgot to take this into account. let's not delete anything

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: `wallets` is every wallet from legacy, right?
I believe simply checking if the wallet is already present in the new prefix is enough to do the skip (i.e. means it's already migrated, skip it to avoid overwriting); no need to check the cache for success

migrations/001-config-s3-structure/migrate-cli.ts
sidvishnoi: They're specific to each migration.

migrations/001-config-s3-structure/__tests__/index.test.ts
sidvishnoi, line 1: Don't forget to rename this to `index.test.ts` then 😄

migrations/001-config-s3-structure/index.ts
```diff
@@ -0,0 +1,190 @@
...
+          'dry-run',
+          'd',
+        ])
+        const unknown = Object.keys(opts).filter((k) => !KNOWN_FLAGS.has(k))
+        if (unknown.length > 0) {
+          console.error(
+            `Unknown flag(s): ${unknown.map((k) => `--${k}`).join(', ')}`,
+          )
+          process.exit(1)
+        }
```
sidvishnoi: https://github.com/lukeed/sade?tab=readme-ov-file#optsunknown

migrations/001-config-s3-structure/migrate-cli.ts
sidvishnoi: We can avoid checking if exists if already in log file though. Like subsequent `--batch` runs will only run on the failed ones.

migrations/001-config-s3-structure/migrate-cli.ts
sidvishnoi: Also, delete by prefix  (or batch delete) is much more efficient then deleting individually. We can just run once.

migrations/001-config-s3-structure/migration-log.ts
```diff
@@ -50,6 +41,15 @@ export class MigrationLog {
     return this.failed.length > 0
   }
 
+  get cachedSuccessful(): Set<string> {
```
sidvishnoi: How is this used?

migrations/001-config-s3-structure/migrate-cli.ts
sidvishnoi: Also, if you don't want to delete but still get same benefits, move the file into a separate prefix (like, copy and then delete original).

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: added the check to avoid if exists check later. Also I'd keep `existsAt` check.
If we remove `existsAt` and rely solely on the log, a --wallet run has no guard.
And also if the log gets deleted between runs, the next batch would blindly putJson over everything.

The cost of keeping it is one extra s3 HEAD request per wallet, but only for wallets the log didn't already skip.

migrations/001-config-s3-structure/migration-log.ts
DarianM: To do the check in the log file:D, renamed the getter tho

migrations/001-config-s3-structure/migrate-cli.ts
DarianM: Let's keep it simple then. Delete all with one run after everything is successful

migrations/001-config-s3-structure/index.ts
```diff
@@ -0,0 +1,182 @@
...
+  const wallets = (await s3.listByPrefix(LEGACY_PREFIX)).map(keyToWalletAddress)
+  if (wallets.length === 0) {
+    console.log('\n x No wallets found in legacy prefix')
+    return
+  }
+
+  console.log(`\n Starting batch migration for ${wallets.length} wallets...`)
+
+  for (const wallet of wallets) {
+    if (log.migratedSuccessful.has(wallet)) {
```
sidvishnoi: This getter will read/parse file on each call.

sidvishnoi (approved):
Looks good to me. Let's do a screen share when you run this migration.

migrations/001-config-s3-structure/index.ts
DarianM: oh. that's dumb

sidvishnoi (approved):

PR #651, opened 2026-04-02
feat(frontend): set up Umami for events data collection

### LINKED ISSUE #624: Infrastructure Integration
Get the chosen analytics technology wired into the codebase and environment.

DarianM:
## Summary

Closes #624 

- Adds Umami as a self-hosted analytics solution via a script tag in `root.tsx`, injected only when `UMAMI_HOST` and `UMAMI_WEBSITE_ID` are set
  - Local dev - set vars -> script loads, data goes to the local docker 
  - Prod - set both vars in the worker, data does to production Umami
  - Any env without the vars set - script tag isn't rendered at all, zero analytics (i.e. no data pollution in staging)
- Adds a local dev Umami instance via `localenv/umami/docker-compose.yml`

## Event tracked as setup

| Event | Route |
|---|---|
| `wallet_connected` | `/banner`, `/widget`, `/offerwall` |
| `click_card_tool` | `link` |

## Notes

- Umami automatically tracks page views for every route after the script tag loads


frontend/app/components/redesign/components/ToolsWalletAddress.tsx
```diff
@@ -63,6 +65,7 @@ export const ToolsWalletAddress = ({
       const walletAddressInfo = await getWalletAddress(walletAddressUrl)
       walletActions.setWalletAddressId(walletAddressInfo.id)
       await connect()
+      track('wallet_connected', { wallet_address: snap.walletAddress })
```
sidvishnoi: Please don't do this ;)

frontend/app/lib/umami.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const track = useCallback(
+    (eventName: string, eventData?: Record<string, unknown>) => {
+      window.umami?.track(eventName, { tool, ...eventData })
+    },
+    [tool],
+  )
+  return <TrackContext.Provider value={track}>{children}</TrackContext.Provider>
+}
+
+export function useTrack(): TrackFn {
```
sidvishnoi:
```suggestion
export function useTrackEvent(): TrackFn {
```

frontend/app/lib/umami.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type TrackFn = (eventName: string, eventData?: Record<string, unknown>) => void
+
+const TrackContext = createContext<TrackFn>(() => {})
+
+export function TrackProvider({
+  tool,
+  children,
+}: {
+  tool: Tool
```
sidvishnoi: What if we want to track something on tools home page, or on other tools that aren't in `Tool`?

frontend/app/routes/offerwall.tsx
```diff
@@ -142,6 +144,7 @@ export default function Offerwall() {
   }
 
   return (
+    <TrackProvider tool={TOOL_OFFERWALL}>
```
sidvishnoi: I think these should be in `root.tsx` only. Supporting a default "tool" value is only a good to have, if compared to tracking all pages consistently.

localenv/umami/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    healthcheck:
+      test: ['CMD-SHELL', 'pg_isready -U umami']
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+  umami:
+    image: ghcr.io/umami-software/umami:postgresql-latest
+    ports:
+      - '3001:3000'
```
sidvishnoi: Consider using a different port, closer to other localenv ports (not close to app ports).

.env.sample
```diff
@@ -11,3 +11,7 @@ AWS_ACCESS_KEY_ID="AWS_KEY_ID"
 AWS_SECRET_ACCESS_KEY="AWS_SECRET_KEY"
 AWS_S3_ENDPOINT="http://localhost:8081" # for dev
 # AWS_S3_ENDPOINT="https://bucket-name.s3.bucket-region.amazonaws.com" # when using AWS
+
+# Umami Analytics
+UMAMI_HOST="http://localhost:3001" # for dev
+UMAMI_WEBSITE_ID="<your-website-uuid>"
```
sidvishnoi: Any reason to use env vars compared to build time constants depending on environment?

The build script can still use env vars when they're available:
https://github.com/interledger/publisher-tools/blob/fe2045169c643ef2715c2fa069db11d244687114/shared/defines/index.ts#L1
https://github.com/interledger/publisher-tools/blob/fe2045169c643ef2715c2fa069db11d244687114/frontend/vite.config.ts#L11

DarianM: use env vars because of `UMAMI_WEBSITE_ID="<your-website-uuid>"` , we cannot have the user change this in defines index for local dev

frontend/app/lib/umami.tsx
DarianM: right. I would avoid passing the tool every time I want to track something redundant, like the profile save event for widget, banner, offerwall.
I added a new event to seehow this would look now

.env.sample
DarianM: ,but i managed a solution using `--env-file` flag

frontend/app/components/redesign/components/ToolsWalletAddress.tsx
```diff
@@ -27,7 +27,7 @@ export const ToolsWalletAddress = ({
 }: Props) => {
   const { connect, disconnect } = useConnectWallet(snap, walletActions)
   const uiActions = useUIActions()
-  const track = useTrackEvent()
+  const event = useTrackEvent()
```
sidvishnoi: Let's be more explicit and call it `trackEvent`?

frontend/app/routes/_index.tsx
```diff
@@ -184,7 +187,10 @@ export default function Index() {
             }}
           >
             {setupTools.map((tool, index) => (
-              <li key={index}>
+              <li
+                key={index}
+                onClick={() => event('click_card_tool', { link: tool.link })}
```
sidvishnoi: Maybe add this on `<ToolCard>`, as that's a link (`<a>`)?

frontend/app/root.tsx
```diff
@@ -6,31 +6,47 @@ import {
...
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <Meta />
         <Links />
+        {UMAMI_HOST && UMAMI_WEBSITE_ID && (
+          <script
+            defer
+            src={`${UMAMI_HOST}/script.js`}
+            data-website-id={UMAMI_WEBSITE_ID}
+          />
+        )}
```
kjmitchelljr, line 43: nit: with this script injection here, do you think we should do any checks to make sure the URLs aren't malformed?

frontend/app/components/redesign/components/ToolsWalletAddress.tsx
```diff
@@ -63,6 +65,7 @@ export const ToolsWalletAddress = ({
       const walletAddressInfo = await getWalletAddress(walletAddressUrl)
       walletActions.setWalletAddressId(walletAddressInfo.id)
       await connect()
+      trackEvent('wallet_connected')
```
kjmitchelljr, line 68: nit: could be a later add, but thinking keeping the events in a separate file with a const map of event names as more events get added

kjmitchelljr (approved):
Looks good, agree with the feedback Sid gave, but see those changes have already been added 👍🏿

docs/env-vars.md
```diff
@@ -123,6 +125,22 @@ AWS_S3_ENDPOINT="https://your-bucket-name.s3.your-region.amazonaws.com"
 
 </details>
 
+### Umami Analytics
+
+Optional. If unset, the analytics script is silently skipped.
+For local dev, start the Umami instance first:
+
+```sh
+cd localenv/umami && docker compose up -d
```
sidvishnoi: Can we add a `package.json` with a `start` script that does this? Then each service can just be started with start/dev.

frontend/app/lib/umami.tsx
sidvishnoi, line 1: Rename this file to `analytics`, so if we swap umami in future, the interface/imports stay same.

frontend/app/lib/umami.tsx
```diff
@@ -0,0 +1,27 @@
+import { createContext, useCallback, useContext } from 'react'
+import type { ReactNode } from 'react'
+import type { Tool } from '@shared/types'
+
+type TrackFn = (eventName: string, eventData?: Record<string, unknown>) => void
```
sidvishnoi: This way, we'll exclude arrays and nested objects that don't work the way we wish they would.
```suggestion
type TrackFn = (eventName: string, eventData?: Record<string, string 
 number | boolean | null>) => void
```

frontend/app/lib/umami.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export function TelemetryProvider({
+  tool,
+  children,
+}: {
+  tool?: Tool
+  children: ReactNode
+}) {
+  const track = useCallback(
+    (eventName: string, eventData?: Record<string, unknown>) => {
```
sidvishnoi: Does this work?
```suggestion
  const track = useCallback<TrackFn>(
    (eventName, eventData) => {
```

frontend/app/root.tsx
```diff
@@ -6,31 +6,47 @@ import {
...
 import faviconSvg from '~/assets/images/favicon.svg?url'
+import { TelemetryProvider } from '~/lib/umami'
 import { UIProvider } from '~/stores/uiStore'
 import stylesheet from '~/tailwind.css?url'
 import { XCircle } from './components/icons.js'
 import { Button } from './components/index.js'
 
 export default function App() {
+  const { pathname } = useLocation()
+  const tool = TOOLS.find((t) => pathname.startsWith(`/${t}`))
```
sidvishnoi: Won't this force a full path rerender on location change, even if some part of it stays as is? Would it be better to handle this within TelemetryProvider?

frontend/app/root.tsx
```diff
@@ -6,31 +6,47 @@ import {
...
       <head>
         <meta charSet="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <Meta />
         <Links />
+        {UMAMI_HOST && UMAMI_WEBSITE_ID && (
+          <script
+            defer
+            src={`${UMAMI_HOST}/script.js`}
+            data-website-id={UMAMI_WEBSITE_ID}
```
sidvishnoi, line 40: Should we add `data-domains` attribute too, so in production, we only track when on right domain (and not on PR previews/staging)?

DarianM: I noticed. web mon sets `data-domains` 
Any env without the vars set - script tag isn't rendered at all, zero analytics (i.e. no data pollution in staging).
I think we're safe in our case. set vars in local dev and prod only

sidvishnoi: When we set up these vars in GitHub (CI), they'll end up setting for both staging/previews and production.

DarianM: right,  then we need to specify `data-domain` to `localhost` for dev and `webmonetization.org` prod. likely through another variable

package.json
```diff
@@ -16,7 +16,9 @@
     "lint": "eslint . --fix --cache --cache-location 'node_modules/.cache/eslintcache'",
     "lint:check": "eslint . --cache --cache-location 'node_modules/.cache/eslintcache'",
     "typecheck": "pnpm -r typecheck",
-    "test": "pnpm -r test"
+    "test": "pnpm -r test",
+    "analytics:start": "pnpm --filter localenv-analytics start",
+    "analytics:stop": "pnpm --filter localenv-analytics stop"
```
sidvishnoi: These can be replaced with `pnpm -C localenv/analytics start` command. If you believe that additional scripts like these are more ergonomic, feel free to keep them.

localenv/analytics/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "localenv-analytics",
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "start": "docker compose up -d",
```
sidvishnoi: Do you want to update `.vscode/tasks.json` to run analytics as well? It might be ok if it crashes - or we can run conditionally based on env var file perhaps?

sidvishnoi (approved):

localenv/analytics/package.json
DarianM: I would avoid adding more shell conditionals that are maybe fragile cross platform for this case.
if it crashes, it crashes.

sidvishnoi: I noticed the other commands use `dev` instead of start..

frontend/app/root.tsx
```diff
@@ -25,12 +27,25 @@ export default function App() {
...
+            data-domains={[
+              new URL(APP_URL.production).hostname,
+              new URL(APP_URL.development).hostname,
```
sidvishnoi: Maybe only add development one during dev, if not too difficult to filter here.

DarianM: changed, added a check for is development; let's see how that would fit

localenv/analytics/package.json
DarianM: changed the naming, to fit `dev` and `down`

sidvishnoi (approved):

sidvishnoi (approved):

PR #656, opened 2026-04-07
refactor(widget): centralize interaction view state in InteractionView

### LINKED ISSUE #481: [refactor] improve interaction view state in the Widget tool component
## Problem
The  payment interaction view  component in from the Widget tool  `/views/interaction/interaction.ts` manages multiple view states (`authorizing`, `processing`, `success`, `failed`) with separate render methods for each. 

has several issues:
- view rendering in spread across multiple separate methods
- inconsistent CTA methods to buttons
- difficult to maintain, adding new states or modifying requires changes across multiple methods 

#### Goals

- [ ] create a unified view rendering system for payment interaction views
- [ ] reduce duplication and centralize view state management (icons, title, description, buttons)

kjmitchelljr:
## Summary

Refactors the widget **payment interaction** screen so all views share one rendering path

## What changed

- **`resolveInteractionView()`** maps `currentView` to correct view
- **`renderInteractionView(view)`** renders the shared layout once

Closes #481 

## Follow-up: automated tests

I'm still working to add automated tests for this flow, but as I was going through I wondered what would be best approach. Adding unit tests to perform in isolation or begin to build out E2E tests to test out whole payment flow which may be too much overhead at this point. But could have a starting point


components/src/widget/views/interaction/interaction.ts
```diff
@@ -243,108 +260,99 @@ export class PaymentInteraction extends LitElement {
...
-        return this.renderProcessingView()
-      case 'success':
-        return this.renderSuccessView()
-      case 'failed':
-        return this.renderFailedView()
-      case 'authorizing':
-      default:
-        return this.renderAuthorizingView()
-    }
+    return this.renderInteractionView(this.resolveInteractionView())
```
DarianM: Let’s avoid the nested call. Instead, we can remove `renderInteractionView` entirely and return the HTML directly here. We can destructure everything we need from the switch resolver.

```suggestion
    render() {
  const { titleClass, title, description, image, button } = this.interactionView
  return html`...`
}
```
rest lgtm

DarianM (changes requested):

DarianM (approved):

PR #660, opened 2026-04-14
fix(frontend): vars naming mismatch through build pipeline

DarianM:
Related to #651 
Drop the `BUILD_` prefix convention for vars never injected by clf at runtime. 
These vars are purely build-time, so there's no collision risk or reason to rename them with `BUILD_`

### DIFF
--- a/.github/workflows/deploy-worker/action.yml
+++ b/.github/workflows/deploy-worker/action.yml
@@ -44,10 +44,10 @@ inputs:
   BUILD_AWS_PREFIX:
     description: 'AWS_PREFIX used for config storage'
     required: true
-  BUILD_UMAMI_HOST:
+  UMAMI_HOST:
     description: 'Umami analytics host URL'
     required: false
-  BUILD_UMAMI_WEBSITE_ID:
+  UMAMI_WEBSITE_ID:
     description: 'Umami analytics website ID'
     required: false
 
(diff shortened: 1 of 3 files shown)


sidvishnoi:
The reason they had build_ prefix was to tell us, that these get defined/embedded at build time (compared to runtime), not to prevent clashes. As in, only be adding "public" data (not secrets) in these as these will be embedded in code. 

Can you tell where it caused issues?

DarianM:
> Can you tell where it caused issues?

check this [commit](https://github.com/interledger/publisher-tools/pull/660/changes/bbac584aebf4a5d7a53084c0a18a69611ee5d833) change in vite config,
After, the workflow would correctly map them to `BUILD_UMAMI_HOST` / `BUILD_UMAMI_WEBSITE_ID` env vars during the build step.

Vite just wasn't reading those `BUILD_` prefixed names, but the .dev.vars keys would need to be `BUILD_` now too for local

DarianM:
So since these vars are purely build-time, and not used in clf there's no collision risk, so to keep it clean and easy a solution was to drop the `_BUILD` prefix here.
they are still defined with prefix in vite config

sidvishnoi:
Maybe this was a better approach:
```ts
    BUILD_UMAMI_HOST: JSON.stringify(process.env.BUILD_UMAMI_HOST || process.env.UMAMI_HOST),
```

Issue #625, opened 2026-03-18
Instrument Analytics Events


RabebOthmani:
Capture the events defined in the analytics spec.

PR #664, opened 2026-04-15
feat(frontend): track more events, shared event types

### LINKED ISSUE #663: Implement tools analytics instrumentation
## Summary
Follow-up to #624 (building on #651) to add tool events for analytic tracking

## Scope
- Complete remaining `tools_*` event wiring in frontend

## Expected Results
- [ ] Expected `tools_*` events fire at correct trigger points
- [ ] Lint/typecheck pass

kjmitchelljr:
## Summary
                                                                                                                                                                                                         
  Wires Umami event tracking into the Tools frontend (closes #663, part of #625)

`useTrackEvent` hook publishes a typed event registry                                                                                                                                                              
                                                                                                                                                                                                         
  ## Events tracked                                                                                                                                                                                      
                                                                                                                                                                                                         
  | Event | Triggered from | Payload |
  |---|---|---|
  | `click_card_tool` | landing tool cards | `{ link, tool? }` |
  | `wallet_connected` | `ToolsWalletAddress` connect | `{ wallet_provider, tool }` |                                                                                                                    
  | `wallet_disconnected` | `ToolsWalletAddress` disconnect | `{ tool }` |                                                                                                                               
  | `${tool}_profile_saved` | `useSaveProfile` save | `{ tool }` |                                                                                                                                       
  | `${tool}_script_generated` | `useSaveProfile` script generation | `{ tool }` |                                                                                                                       
  | `link_tag_generated` | `LinkTagGenerator` submit | `{ wallet_provider }` |


frontend/app/hooks/useSaveProfile.tsx
```diff
@@ -48,10 +51,13 @@ export const useSaveProfile = (wallet: WalletStore) => {
 
         if (result.success) {
           actions.commitProfile()
+          trackEvent(TOOLS_EVENTS.SETTINGS_CHANGED)
```
kjmitchelljr: Do we want to go ahead and track field changes?

frontend/app/routes/_index.tsx
```diff
@@ -137,6 +140,11 @@ export const meta: MetaFunction = () => {
 }
 
 export default function Index() {
+  const trackEvent = useTrackEvent()
+
+  useEffect(() => {
+    trackEvent(TOOLS_EVENTS.VIEW_MAIN_PAGE)
+  }, [trackEvent])
```
kjmitchelljr: Is this still needed or would we have those analytics already based on referrers/visited pages?

frontend/app/routes/_index.tsx
```diff
@@ -137,6 +140,11 @@ export const meta: MetaFunction = () => {
 }
 
 export default function Index() {
+  const trackEvent = useTrackEvent()
+
+  useEffect(() => {
+    trackEvent(TOOLS_EVENTS.VIEW_MAIN_PAGE)
```
DarianM: Pageviews are tracked automatically out of the box, no need to add this track event
https://docs.umami.is/docs/tracker-configuration#data-auto-track-v200

sidvishnoi (changes requested):
Consider creating a map (like object) of event name and its expected custom event data (can be undefined/never).
Then instead of exporting each constant, we just let our frontend analytics lib know the possible events, and when we call trackEvent, it'll be correctly typed (and we get to keep all events info in one place too).
Like:
```ts
type Events = {
  eventA: undefined,
  eventB: { tool: ToolName, walletProvider: string }
}
```

Also instead of shared package, start by adding it in frontend only (where do you expect frontend event map to be used for it to be a shared package?)

frontend/app/hooks/useSaveProfile.tsx
DarianM: I’d approach it like this: on profile save, we track which fields have changed compared to the last known state. That way, we avoid adding tracking events to every individual style customization property and can instead handle everything in one place.

Another consideration is to avoid tracking every styling change made during the “build” step before the profile is saved, since that could introduce unnecessary noise and bloat.

kjmitchelljr:
> Consider creating a map (like object) of event name and its expected custom event data (can be undefined/never). Then instead of exporting each constant, we just let our frontend analytics lib know the possible events, and when we call trackEvent, it'll be correctly typed (and we get to keep all events info in one place too). Like:
> 
> ```ts
> type Events = {
>   eventA: undefined,
>   eventB: { tool: ToolName, walletProvider: string }
> }
> ```
> 
> Also instead of shared package, start by adding it in frontend only (where do you expect frontend event map to be used for it to be a shared package?)

Can do. I don't expect frontend event map to be used, but based on the doc I know I'll be soon adding events in CDN and API as well. The thinking was to have it shared across there, however could keep it all separate.

frontend/app/hooks/useSaveProfile.tsx
kjmitchelljr: Sounds good. I'll make that update.

sidvishnoi:
Given CDN & frontend won't be sharing event names (like CDN ones will have a prefix for sure), I don't think they need to be in a shared package.

I'm not sure we need Umami events in API - we've different kind of technical instrumentation there from Cloudflare already; and as CDN/embed is calling the API, we'll have user-facing events from embeds, not "API called" events.

DarianM:
> I'm not sure we need Umami events in API -

Are you suggesting we shouldn’t use Umami in the API for things like tool_embed_render? If so, that could mean our events are no longer in one  place dashboard.
I thought the right place to track tool render is in our api, on GET api/profile

kjmitchelljr:
> > I'm not sure we need Umami events in API -
> 
> Are you suggesting we shouldn’t use Umami in the API for things like tool_embed_render? If so, that could mean our events are no longer in one place dashboard. I thought the right place to track tool render is in our api, on GET api/profile

@sidvishnoi I was basing this off the [doc](https://docs.google.com/document/d/1jpDH2vbxOb1TWCuKBTZTU2LQbOjuExkoXv5G75Smm9o/edit?tab=t.0) where it mentions Count GET /profile{tool} on /api module per what Darian was saying above. Could you help me understand?

sidvishnoi:
Can we discuss tracking the API when we come to it, and here focus on the frontend?

frontend/app/lib/analytics-events.ts
```diff
@@ -0,0 +1,11 @@
+import type { Tool } from '@shared/types'
+
+export type ToolsEventMap = {
```
sidvishnoi: Why is there an `tools_` prefix for each event? If we're to track CDN, we should add prefix there. `frontend` is the primary tracking target, let's not make analytics harder to use for this.

frontend/app/stores/banner-store.ts
```diff
@@ -118,11 +118,23 @@ export const actions = {
...
   commitProfile() {
     const profile = snapshot(banner.profile)
+    const prev = snapshots.get(toolState.activeTab)
+
+    const changedFields = prev
+      ? Object.keys(profile).filter(
+          (k) =>
+            JSON.stringify(profile[k as keyof BannerProfile]) !==
+            JSON.stringify(prev[k as keyof BannerProfile]),
+        )
```
sidvishnoi: We've nested objects. If `font.name` is changed, we'll mark `font` as changed - we'll lose info if `font.name` was changed or `font.size`.

frontend/app/stores/banner-store.ts
```diff
@@ -118,11 +118,23 @@ export const actions = {
...
+      : Object.keys(profile)
```
sidvishnoi: Given this is used in all the tools, extract this into a helper function.

frontend/app/hooks/useSaveProfile.tsx
```diff
@@ -47,11 +49,24 @@ export const useSaveProfile = (wallet: WalletStore) => {
         }
 
         if (result.success) {
-          actions.commitProfile()
+          const changedFields = actions.commitProfile()
+
+          if (changedFields.length > 0) {
+            trackEvent('tools_settings_changed', {
+              tool: toolState.currentToolType,
+              changed_fields: changedFields,
```
sidvishnoi: This will set `changed_fields: string[]`. In Umami, arrays are serialized intro string. We won't be able to analyze this.

Try running Umami locally (`pnpm analytics:start`), and see how the events are passed, and whether they're analyzable.

Consider this approach instead (that's why I think changed field tracking should be a separate PR in itself):
```ts
trackEvent('settings_changed', {
  tool,
  'field.font.name': true, // true means field changed

  //// false means field not changed; so we don't need need to send it to event
  // 'field.font.size': false,

  // if title.text was changed, what's the length of it?
  // It can be better than boolean, as we'll know if people typically write
  // shorter or longer titles; while making sure we're not tracking 
  // exact values (privacy risks).
  'field.title.text': 50,
})
```

sidvishnoi (changes requested):

frontend/app/lib/analytics-events.ts
kjmitchelljr: Has been updated.

frontend/app/hooks/useSaveProfile.tsx
```diff
@@ -48,10 +50,18 @@ export const useSaveProfile = (wallet: WalletStore) => {
...
+          trackEvent('settings_changed', {
+            tool: toolState.currentToolType,
+          })
           if (action === 'script') {
+            trackEvent('script_generated', {
+              tool: toolState.currentToolType,
+            })
             openDialog(<ScriptDialog wallet={wallet} />)
           } else {
+            trackEvent('profile_saved', {
```
sidvishnoi: Do we need two separate events for this? What if we pass an `action` in event data to tell whether it was save or script?

frontend/app/components/redesign/components/LinkTagGenerator.tsx
```diff
@@ -36,6 +38,7 @@ export const LinkTagGenerator = () => {
         const validatedPointer = await validateAndConfirmPointer(pointerInput)
         setParsedLinkTag(htmlEncodePointer(validatedPointer))
         setShowCodeBox(true)
+        trackEvent('generated_tag', { tag_type: 'link_tag' })
```
sidvishnoi: Can we use the same _save_ /_submit_ event here from other tools and pass `tool: "link_tag"` with it?

frontend/app/hooks/useSaveProfile.tsx
```diff
@@ -48,10 +50,18 @@ export const useSaveProfile = (wallet: WalletStore) => {
 
         if (result.success) {
           actions.commitProfile()
-
+          trackEvent('settings_changed', {
```
sidvishnoi: Let's remove this event for now - this one here is fired even if settings aren't changed.

sidvishnoi (changes requested):

frontend/app/hooks/useSaveProfile.tsx
kjmitchelljr: @sidvishnoi think keeping separate would be ideal for now. I was reading https://github.com/umami-software/umami/issues/3123 as it relates to event filtering. From a technical standpoint I'd agree with combining where we can for maintenance. From a stakeholder standpoint and navigating the dashboard keeping separate is best.

frontend/app/components/redesign/components/LinkTagGenerator.tsx
kjmitchelljr: @sidvishnoi would you be able to explain more? I'm also leaning keeping this separate as the link tag generation is independent of tool(s). I would argue that it would make sense include wallet provider which I've added in the next commit.

frontend/app/hooks/useSaveProfile.tsx
sidvishnoi: In this particular case, we do get a distribution by "action" in dashboard. But yes, let's use separate events until Umami has filtering by event properties in dashboard.

frontend/app/components/redesign/components/LinkTagGenerator.tsx
sidvishnoi: I was thinking like we'll keep a "saved" or "submitted" event, and the "tool" event property will change with it. Then we can filter by tool type as needed. But as per your other comment, filtering by event data isn't very good in umami dashboard as of yet. 
So I think we should add tool name to events now, like "generate_{tool}" as otherwise profile saved won't be filterable in itself.

frontend/app/lib/analytics-events.ts
```diff
@@ -0,0 +1,13 @@
...
+
+type ProfileSavedEvent = `${Tool}_profile_saved`
+type ScriptGeneratedEvent = `${Tool}_script_generated`
+
+export type ToolsEventMap = {
+  click_card_tool: { link: string }
+  wallet_connected: { wallet_provider: string }
+  wallet_disconnected: undefined
+  link_tag_generated: { wallet_provider: string }
+} & {
```
sidvishnoi: This `&` feels like a band-aid (see how the types appear in VSCode etc.)

Let's consider alternate approaches:
```ts
type EventRegistry =
  | [event: "click_card_tool", payload: { link: string }]
  | [event: "wallet_connected", payload: { wallet_provider: string }]
  | [event: "wallet_disconnected"]
  | [event: "link_tag_generated", payload: { wallet_provider: string }]
  | [event: `${Tool}_profile_saved`]
  | [event: `${Tool}_script_generated`]

export type TrackFn = (...args: EventRegistry) => void;
export const trackFn: TrackFn = (...args) => {
  const [event, data] = args;
  return;
};
// now `trackFn` shows us all the overload signatures
```

Or:
```ts
type EventRegistry =
  | { event: "click_card_tool"; payload: { link: string } }
  | { event: "wallet_connected"; payload: { wallet_provider: string } }
  | { event: "wallet_disconnected"; payload: never }
  | { event: "link_tag_generated"; payload: { wallet_provider: string } }
  | { event: `${Tool}_profile_saved`; payload: never }
  | { event: `${Tool}_script_generated`; payload: never };

export type ToolsEventMap = {
  [E in EventRegistry as E["event"]]: E["payload"];
};

export type TrackFn = <E extends EventRegistry["event"]>(
  ...args: ToolsEventMap[E] extends never
    ? [eventName: E]
    : [eventName: E, data: ToolsEventMap[E]]
) => void;

export const trackFn: TrackFn = (...args) => {
  const [event, payload] = args;
  return;
};
```

frontend/app/lib/analytics.tsx
```diff
@@ -14,12 +16,13 @@ export function TelemetryProvider({ children }: { children: ReactNode }) {
...
   const tool = TOOLS.find((t) => pathname.startsWith(`/${t}`))
 
-  const track = useCallback<TrackFn>(
-    (eventName, eventData) => {
-      window.umami?.track(eventName, tool ? { tool, ...eventData } : eventData)
-    },
+  const track = useCallback(
+    ((eventName, eventData) => {
+      window.umami?.track(eventName, { ...(tool && { tool }), ...eventData })
+    }) as TrackFn,
```
sidvishnoi: Let's avoid `as` unless necessary.

sidvishnoi (approved):

PR #675, opened 2026-04-20
fix(api): ensure `AWS_PREFIX` is passed at build-time

sidvishnoi:


api/package.json
```diff
@@ -4,10 +4,10 @@
   "scripts": {
     "predev": "node -e \"require('fs').copyFileSync('../.dev.vars', '.dev.vars');\"",
     "dev": "wrangler dev --persist-to ../.wrangler",
-    "build": "wrangler deploy --dry-run --outdir dist",
+    "build": "exit 0",
```
sidvishnoi: `wrangler deploy` does it for us. Removed the script to avoid duplicating build command.

api/package.json
```diff
@@ -4,10 +4,10 @@
...
     "preview": "wrangler dev",
     "typecheck": "tsc --noEmit",
-    "deploy": "wrangler deploy"
+    "deploy": "wrangler deploy --define BUILD_AWS_PREFIX=\"$BUILD_AWS_PREFIX\""
```
sidvishnoi: Given we deploy from GitHub CI (Linux), we don't need to make it Windows-safe here. One can deploy safely from git-bash otherwise.

sidvishnoi: Not sure if we need more escaping on this..

api/package.json
DarianM: do we need to keep the `build` here? can't remove entirely

sidvishnoi: If we run `pnpm -r build`, then having the script that returns success can be helpful - if it wasn't there, the `pnpm -r build` can fail (unless specific flags are passed).

DarianM (approved):

PR #679, opened 2026-04-22
fix(ci/deploy): use `environment` for deployments

sidvishnoi:


DarianM (approved):
can we continue with cleaning up the variables

PR #680, opened 2026-04-22
chore(ci/setup): update dependencies, action setup

sidvishnoi:
Based on https://github.com/actions/setup-node/blob/v6.4.0/docs/advanced-usage.md#caching-packages-data

We were using v2 of pnpm-setup. Saw a warning in past runs, as it uses Node.js v20.

### DIFF
--- a/.github/workflows/setup/action.yml
+++ b/.github/workflows/setup/action.yml
@@ -1,4 +1,4 @@
-name: 'WMTools Environment Setup'
+name: 'Publisher Tools Environment Setup'
 description: 'Installs NodeJS, PNPM, dependencies and PNPM restores cache'
 
 runs:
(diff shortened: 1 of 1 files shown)


.github/workflows/setup/action.yml
```diff
@@ -1,34 +1,18 @@
...
-        path: ${{ steps.pnpm-cache.outputs.pnpm_cache_dir }}
-        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+    - name: Setup NodeJS
+      uses: actions/setup-node@v6
+      with: { node-version-file: .nvmrc, cache: 'pnpm' }
 
     - name: Install dependencies
       shell: bash
-      run: pnpm install --frozen-lockfile
+      run: pnpm install
```
kjmitchelljr, line 18: Could you explain the reason behind this change? From my understanding `pnpm install --frozen-lockfile` should remain or even possibly update to `pnpm ci`

sidvishnoi: --frozen-lockfile is set by default by pnpm in CI environments.

kjmitchelljr (approved):

PR #681, opened 2026-04-22
feat(api/wrangler.toml): enable worker observability

sidvishnoi:
This will need a deploy (release) to come into effect.


kjmitchelljr (approved):
Straightforward change.

Issue #682, opened 2026-04-23
Implement analytics instrumentation for embedded tools (Banner & Offerwall


kjmitchelljr:
## Summary
Follow-up to [#663](https://github.com/interledger/publisher-tools/issues/663) to specifically add analytics event tracking for embedded components, such as the **Banner** and **Offerwall**.

## Scope
* Implement analytics event wiring for the **Banner** component
* Implement analytics event wiring for the **Offerwall** component

## Expected Results
- [ ] Banner-specific analytics events fire at correct trigger points
- [ ] Offerwall-specific analytics events fire at correct trigger points
- [ ] Lint/typecheck pass

PR #683, opened 2026-04-23
feat(analytics): base event tracking for banner and offerwall embeds

kjmitchelljr:
## Summary

  Adds `embed.{tool}.click_extension_link` event tracking when visitors click "Install the Web Monetization Extension" in the banner or offerwall embeds

  ## Architecture
  - **CDN** (`navigator.sendBeacon`) → `POST /events` on the API
  - **API** validates the event name (`embed.*` prefix), forwards to Umami's `/api/send` with publisher's geo headers (per [Umami's CF docs](https://docs.umami.is/docs/enable-cloudflare-headers))
  - **Schema** ([`api/src/routes/events.ts`](api/src/routes/events.ts)) exports `EventPayload` (inner payload) and
  `EventBody` (full envelope); CDN imports both so request bodies stay in sync with the server schema
  - `trackEventFactory(tool)` captures host context once per script entry; returned `trackEvent(name, data?)` takes
  positional args
### Follow Ups
  - Origin allow-list + rate limit on /events

### Part of
#625 / #682

Three small departures from your original:


cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,39 @@
+import { UMAMI_HOST, UMAMI_WEBSITE_ID } from '@shared/defines'
+
+type ExtensionLinkSource = 'banner' | 'offerwall'
+type ClickLinkEvent = `click_link_${ExtensionLinkSource}`
+
+export type CdnEventMap = {
+  [K in ClickLinkEvent]: undefined
```
kjmitchelljr: Should the event payload include `{ link: string }` (the resolved extension store URL) to allow slicing by browser/store in the dashboard or is the tracking from Posthog fine on this front?

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,39 @@
...
+export function injectUmami() {
+  if (!UMAMI_HOST || !UMAMI_WEBSITE_ID) return
+  if (typeof window === 'undefined' || typeof document === 'undefined') return
+  if (window.umami || umamiInjected) return
+
+  umamiInjected = true
+  const script = document.createElement('script')
+  script.defer = true
+  script.src = `${UMAMI_HOST}/script.js`
+  script.setAttribute('data-website-id', UMAMI_WEBSITE_ID)
```
sidvishnoi: Will this work on a third party website where our script is embedded?

sidvishnoi (changes requested):
This will end up injecting umami into the website. It'll track the website's page views etc. Also, what if the site already has umami? Which `window.track()` will get used? Also, what if site's CSP doesn't allow unami script/API?

We should instead proxy the events from CDN to go via a custom endpoint in API. That'll mean calling umami API from our API, and not using umami script.

cdn/src/lib/analytics.ts
kjmitchelljr: @sidvishnoi definitely something I want to further test - I was reading something about needing a domain whitelist, but wasn't seeing the option in Umami. I do like the idea of using custom API endpoint would simplify things

api/src/routes/events.ts
```diff
@@ -0,0 +1,41 @@
+import z from 'zod'
+import { app } from '../app.js'
+
+const eventSchema = z.object({
+  name: z.enum(['click_link_banner', 'click_link_offerwall']),
+})
```
sidvishnoi, line 21: Might want to create a `z.union` with all the event types here, similar to frontend, as events might have metadata here as well. Then we can export a `TrackFn` type from here and use it in CDN.

api/src/routes/events.ts
```diff
@@ -0,0 +1,41 @@
...
+    return body(null, 400)
+  }
+
+  try {
+    await fetch(`${env.UMAMI_HOST}/api/send`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'User-Agent': req.header('user-agent') ?? '',
+        'X-Forwarded-For': req.header('cf-connecting-ip') ?? '', // do we need this?
```
sidvishnoi: We might want to pass through most of the headers here, not just these.

api/src/routes/events.ts
```diff
@@ -0,0 +1,41 @@
...
+      },
+      body: JSON.stringify({
+        type: 'event',
+        payload: {
+          website: env.UMAMI_WEBSITE_ID,
+          hostname: HOSTNAME,
+          url: '/',
+          name: event.name,
+        },
```
sidvishnoi: Let's send the entire payload from client only, so that this API endpoint is just a proxy to `UMAMI/api/send`.

api/src/routes/events.ts
```diff
@@ -0,0 +1,41 @@
...
+        'Content-Type': 'application/json',
+        'User-Agent': req.header('user-agent') ?? '',
+        'X-Forwarded-For': req.header('cf-connecting-ip') ?? '', // do we need this?
+      },
+      body: JSON.stringify({
+        type: 'event',
+        payload: {
+          website: env.UMAMI_WEBSITE_ID,
+          hostname: HOSTNAME,
+          url: '/',
```
sidvishnoi: Perhaps we should use a different url here, to denote these are coming from embeds. With `/`, it'll interfere with existing events.

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,9 @@
+import { API_URL } from '@shared/defines'
+
+export type CdnEvent = 'click_link_banner' | 'click_link_offerwall'
+
+export function trackEvent(name: CdnEvent): void {
```
sidvishnoi: Events from here will have metadata too. Let's not limit to event name only.

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,9 @@
+import { API_URL } from '@shared/defines'
+
+export type CdnEvent = 'click_link_banner' | 'click_link_offerwall'
```
sidvishnoi: You can export a type from API, and import it from `publishers-tools-api` package in CDN.

sidvishnoi (changes requested):
Moving in right direction now..

api/src/routes/events.ts
sidvishnoi: Also: add a `embed.` prefix to all the events, so we know they're coming from third-party websites.

kjmitchelljr: > Might want to create a `z.union` with all the event types here, similar to frontend, as events might have metadata here as well. Then we can export a `TrackFn` type from here and use it in CDN.

Done - changed to `z.union` so each event has its own metadata (allows to expand in future if need be). `Trackn` is exported for CDN consumption.

> Also: add a `embed.` prefix to all the events, so we know they're coming from third-party websites.

Embed prefix added

api/src/routes/events.ts
kjmitchelljr: Still debating whether to just pass-through all here. Went with `allowlist` that made sense to me, but we could include more. Let me know your thoughts.

api/src/routes/events.ts
kjmitchelljr: > Let's send the entire payload from client only, so that this API endpoint is just a proxy to `UMAMI/api/send`.

Done - the payload is coming from the CDN

api/src/routes/events.ts
kjmitchelljr: Done - using `/embed/banner` and `/embed/offereall` for the URL

cdn/src/lib/analytics.ts
kjmitchelljr: Done - additional metadatsa has been added with trackEvent taking in `{ name, data? }`

cdn/src/lib/analytics.ts
kjmitchelljr: Done - type is exported from API

api/src/routes/events.ts
```diff
@@ -0,0 +1,62 @@
...
+    try {
+      await fetch(`${env.UMAMI_HOST}/api/send`, {
+        method: 'POST',
+        headers,
+        body: JSON.stringify({
+          ...event,
+          payload: {
+            ...event.payload,
+            website: env.UMAMI_WEBSITE_ID,
+            hostname: env.UMAMI_HOSTNAME,
```
kjmitchelljr: @DarianM @sidvishnoi would you be able to provide clarity on where the environment variables are kept? Would need to have one for `UMAMI_HOSTNAME` if I understand correctly, but please feel free to provide more insight

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,16 @@
+import type { TrackFn } from 'publisher-tools-api'
+import { API_URL } from '@shared/defines'
+
+type TrackArgs = Omit<TrackFn, 'url'>
+
+export function trackEvent(event: TrackArgs): void {
+  if (!API_URL) return
+  // assumes all event names follow 'embed.click_link_<source>'
+  const url = `/embed/${event.name.replace('embed.click_link_', '')}`
```
sidvishnoi: Alternative: How about we use a higher-order function like this:
```ts
export function trackEventFactory(tool: Tool) {
  return (event: TrackArgs) => {
    // ...
  }
}

const trackEvent = trackEventFactory('banner') // at some top-level in each of our tool script in CDN
trackEvent({name, /* ... */ })
```

cdn/src/utils/offerwall/WebMonetizationCustomOfferwallChoice.ts
```diff
@@ -98,8 +99,12 @@ export class WebMonetizationCustomOfferwallChoice implements OfferwallCustomChoi
 
     const owElem = document.createElement(elementName) as OfferwallModal
     const actions = owElem.setController({
-      onExtensionLinkClick() {
-        // can start tracking
+      onExtensionLinkClick(e) {
+        const { link } = (e as CustomEvent<{ link: string }>).detail
```
sidvishnoi: Nit: Let's update the `Controller` interface to denote what the type of `e` is.

api/src/app.ts
```diff
@@ -13,6 +13,9 @@ export type Env = {
   OP_PRIVATE_KEY: string
   OP_KEY_ID: string
   PUBLISHER_TOOLS_KV: KVNamespace
+  UMAMI_HOST?: string
+  UMAMI_WEBSITE_ID?: string
+  UMAMI_HOSTNAME?: string
```
sidvishnoi: These 3 won't be available in env vars, but rather available as build time vars:
https://github.com/interledger/publisher-tools/blob/8e215c8bda156520ac39dacd89570685eefc15ec/api/build.ts#L5-L7

Then, we'd use them via `@shared/defines`:
https://github.com/interledger/publisher-tools/blob/8e215c8bda156520ac39dacd89570685eefc15ec/api/src/index.ts#L1

sidvishnoi: Would need to update CI a bit as well:
https://github.com/interledger/publisher-tools/blob/8e215c8bda156520ac39dacd89570685eefc15ec/.github/workflows/deploy.yml#L140

sidvishnoi: Nit: Maybe call it `UMAMI_API_HOST`

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,16 @@
+import type { TrackFn } from 'publisher-tools-api'
+import { API_URL } from '@shared/defines'
+
+type TrackArgs = Omit<TrackFn, 'url'>
```
sidvishnoi: Perhaps call `TrackFn` as something like: `TrackPayload`?

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,16 @@
...
+export function trackEvent(event: TrackArgs): void {
+  if (!API_URL) return
+  // assumes all event names follow 'embed.click_link_<source>'
+  const url = `/embed/${event.name.replace('embed.click_link_', '')}`
+  // text/plain to avoid CORS preflight on sendBeacon (body is JSON)
+  const blob = new Blob(
+    [JSON.stringify({ type: 'event', payload: { ...event, url } })],
+    { type: 'text/plain' },
+  )
+  navigator.sendBeacon?.(`${API_URL}/events`, blob)
```
sidvishnoi: This gave me a HTTP 400 validation error as I used the PR preview as well as checked out locally.
```json
{
  "error": {
    "message": "Validation failed",
    "code": "VALIDATION_ERROR",
    "details": {
      "issues": [
        {
          "path": "type",
          "message": "Invalid input: expected \"event\"",
          "code": "invalid_value"
        },
        {
          "path": "payload",
          "message": "Invalid input",
          "code": "invalid_union"
        }
      ]
    }
  }
}
```

sidvishnoi (changes requested):
Nearly there, but little broken..

cdn/src/lib/analytics.ts
sidvishnoi: I think it could be because we're sending `text/plain`, but server validator is validating on `zValidator('json', schema`)`?

sidvishnoi: This worked locally:
```diff
diff --git a/api/src/routes/events.ts b/api/src/routes/events.ts
index 582bcfff..163f1278 100644
--- a/api/src/routes/events.ts
+++ b/api/src/routes/events.ts
@@ -24,14 +24,13 @@ export type TrackFn = z.infer<typeof payloadSchema>
 
 app.post(
   '/events',
-  zValidator('json', eventSchema),
   async ({ req, env, body }) => {
+    const event = z.parse(eventSchema, await req.json())
+    
     if (!env.UMAMI_HOST || !env.UMAMI_WEBSITE_ID || !env.UMAMI_HOSTNAME) {
       return body(null, 204)
     }
 
-    const event = req.valid('json')
-
     const headers = new Headers({ 'content-type': 'application/json' })
     for (const h of ['user-agent', 'accept-language', 'referer']) {
       const v = req.header(h)

```

api/src/routes/events.ts
sidvishnoi: Replied here: https://github.com/interledger/publisher-tools/pull/683#discussion_r3201447802


These build time vars are stored in https://github.com/interledger/publisher-tools/settings/variables/actions

api/src/routes/events.ts
sidvishnoi: Umami docs shows these headers: https://docs.umami.is/docs/enable-cloudflare-headers
I think we want to skip the Host as headers for sure, but we do want to pass through the website domain to be available in our event payload/data, so we know which sites are embedding our tools.

cdn/src/lib/analytics.ts
sidvishnoi: This can also set a `host` context: domain/page where the the script is embedded, which can be used on all events from given component.

components/src/banner.ts
```diff
@@ -62,7 +62,13 @@ export class Banner extends LitElement {
   }
 
   private handleLinkClick() {
-    // TODO: do anything other than open the link in a new tab, like analytics, showing some thank you message etc.
+    this.dispatchEvent(
+      new CustomEvent('click-extension-link', {
+        bubbles: true,
+        composed: true,
```
sidvishnoi: Why do we want these two settings?

cdn/src/lib/analytics.ts
kjmitchelljr: Fixed here [f671e58](https://github.com/interledger/publisher-tools/pull/683/commits/f671e5806c2ac2b2d535c9758e9c2f910bcc850d), though I was seeing 500 error on my initial tests so added try/catch to properly get 400 response.

api/src/app.ts
kjmitchelljr: Added in this commit [a1b024a](https://github.com/interledger/publisher-tools/pull/683/commits/a1b024ae0cb7dcaebb29a17621173e9e9fb2b97c), have a better understanding now, thank you

api/src/routes/events.ts
kjmitchelljr: Fixed here [a1b024a](https://github.com/interledger/publisher-tools/pull/683/commits/a1b024ae0cb7dcaebb29a17621173e9e9fb2b97c), but when I navigated to https://github.com/interledger/publisher-tools/settings/variables/actions I only saw `CLOUDFLARE_ACCOUNT_ID` variable. None of the others, unless still need to be added?

cdn/src/utils/offerwall/WebMonetizationCustomOfferwallChoice.ts
kjmitchelljr: Fixed here [903bfe1](https://github.com/interledger/publisher-tools/pull/683/commits/903bfe16e85ed3274a15527f82d75f1056c355fa)

cdn/src/lib/analytics.ts
kjmitchelljr: Changes found here: [1d22a5a](https://github.com/interledger/publisher-tools/pull/683/commits/1d22a5ae997cdec01b5921e531ea7a6523b9bd93)

components/src/banner.ts
kjmitchelljr: Initial reasoning would be following the same practices used throughout the code where `composed: true` also included `bubbles: true`, but upon further review you are correct, these flags wouldn't be needed.

Change added here: [2ce05a5](https://github.com/interledger/publisher-tools/pull/683/commits/2ce05a599a4986e8fabcfd43c51ba2429ef530b1)

api/src/routes/events.ts
kjmitchelljr: Got it - the change was made here: [0aad2d9](https://github.com/interledger/publisher-tools/pull/683/commits/0aad2d9e0256a6e1e79ce26d588ac29f9ffd61b2)

kjmitchelljr:
@sidvishnoi All the changes should be there now, adding in small test for POST, but wanted to ask about the Github Actions variables, right now I'm not seeing UMAMI related vars unless that is something still to add

.github/workflows/deploy-worker/action.yml
```diff
@@ -42,7 +42,10 @@ inputs:
     description: 'AWS_PREFIX used for config storage'
     required: true
   BUILD_UMAMI_HOST:
-    description: 'Umami analytics host URL'
+    description: 'Umami analytics host URL (frontend tracker script)'
+    required: false
+  BUILD_UMAMI_API_HOST:
+    description: 'Umami analytics host URL (server-to-server forward target)'
```
sidvishnoi: Can we get `UMAMI_API_HOST` using existing `UMAMI_HOST` or are they entirely independent? Thinking if we do need a new variable.

api/src/routes/events.ts
```diff
@@ -0,0 +1,71 @@
...
+import { UMAMI_API_HOST, UMAMI_WEBSITE_ID } from '@shared/defines'
+import { app } from '../app.js'
+
+const payloadSchema = z.object({
+  name: z.string().startsWith('embed.'),
+  url: z.string(),
+  data: z.object({
+    hostname: z.string(),
+    link: z.string().optional(),
+  }),
```
sidvishnoi: This restricts event data by a lot - what if we want to pass more data? We'll have to update this?

api/src/routes/events.ts
```diff
@@ -0,0 +1,71 @@
+import z from 'zod'
+import { UMAMI_API_HOST, UMAMI_WEBSITE_ID } from '@shared/defines'
+import { app } from '../app.js'
+
+const payloadSchema = z.object({
+  name: z.string().startsWith('embed.'),
+  url: z.string(),
+  data: z.object({
+    hostname: z.string(),
+    link: z.string().optional(),
```
sidvishnoi: Can you document each of `url`, `hostname`, `link` - which is meant as which, as they sound very similar.

api/src/routes/events.ts
```diff
@@ -0,0 +1,71 @@
...
+  let event: z.infer<typeof eventSchema>
+  try {
+    event = z.parse(eventSchema, await req.json())
+  } catch {
+    return body(null, 400)
+  }
+
+  if (!UMAMI_API_HOST || !UMAMI_WEBSITE_ID) {
+    return body(null, 204)
+  }
```
sidvishnoi: Let's move this as the first check (before we validate payload).

api/src/routes/events.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  UMAMI_WEBSITE_ID: 'test-website-id',
+}))
+
+describe('POST /events', () => {
+  let fetchMock: ReturnType<typeof vi.fn>
+
+  beforeEach(() => {
+    fetchMock = vi.fn().mockResolvedValue(new Response(null, { status: 200 }))
+    globalThis.fetch = fetchMock as unknown as typeof fetch
+  })
```
sidvishnoi, line 20: Clear mock `afterEach` as well?

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,27 @@
...
+
+type TrackArgs = {
+  name: string
+  data?: Omit<TrackPayload['data'], 'hostname'>
+}
+
+export function trackEventFactory(tool: Tool) {
+  const hostname = window.location.hostname
+
+  return ({ name, data }: TrackArgs): void => {
```
sidvishnoi: This would be more ergonomic.
```suggestion
  return (name, data): void => {
```

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,27 @@
...
+    navigator.sendBeacon?.(
+      `${API_URL}/events`,
+      JSON.stringify({
+        type: 'event',
+        payload: {
+          name: `embed.${tool}.${name}`,
+          url: `/embed/${tool}`,
+          data: { hostname, ...data },
+        },
+      }),
```
sidvishnoi: Let's export an `EventsData` like interface from API, so the body structures remain in sync,
```suggestion
      } satisfies EventData),
```

sidvishnoi (changes requested):
Nearly there!

.github/workflows/deploy-worker/action.yml
kjmitchelljr: Yes, you're correct, for some reason I thought I saw somewhere that the API events would be separate from the frontend. But as of right now, they would point to the same Umami cloud instance. I've made the update here: https://github.com/interledger/publisher-tools/pull/683/commits/87b0a5c3dcf1e406a891e871b4c4f50657800e1f

api/src/routes/events.test.ts
kjmitchelljr: Added here [a6063d1](https://github.com/interledger/publisher-tools/pull/683/commits/a6063d1eb718680d9b770da76218806cc33cc73a)

api/src/routes/events.ts
kjmitchelljr: Change made in this commit [a6063d1](https://github.com/interledger/publisher-tools/pull/683/commits/a6063d1eb718680d9b770da76218806cc33cc73a)

api/src/routes/events.ts
kjmitchelljr: Change made in this commit [a6063d1](https://github.com/interledger/publisher-tools/pull/683/commits/a6063d1eb718680d9b770da76218806cc33cc73a)

api/src/routes/events.ts
kjmitchelljr: Change made in this commit [a6063d1](https://github.com/interledger/publisher-tools/pull/683/commits/a6063d1eb718680d9b770da76218806cc33cc73a)

cdn/src/lib/analytics.ts
kjmitchelljr: Updated here: [2876510](https://github.com/interledger/publisher-tools/pull/683/commits/2876510819866bb83185ebf02f474da743694f12)

cdn/src/lib/analytics.ts
kjmitchelljr: Updated here: [2876510](https://github.com/interledger/publisher-tools/pull/683/commits/2876510819866bb83185ebf02f474da743694f12)

api/src/routes/events.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  try {
+    await fetch(`${UMAMI_HOST}/api/send`, {
+      method: 'POST',
+      headers,
+      body: JSON.stringify({
+        ...event,
+        payload: {
+          ...event.payload,
+          website: UMAMI_WEBSITE_ID,
```
sidvishnoi, line 67: Let's also add a [tag](https://docs.umami.is/docs/tags) for events from embeds.
```suggestion
          website: UMAMI_WEBSITE_ID,
          tag: 'embed',
```

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  return (name: string, data?: EventData): void => {
+    if (!API_URL) return
+    const body: EventBody = {
+      type: 'event',
+      payload: {
+        name: `embed.${tool}.${name}`,
+        url: `/embed/${tool}`,
+        data: { hostname, ...data },
+      },
```
sidvishnoi, line 19: Compared to `umami.track()`, we're missing the `screen` size details.
https://docs.umami.is/docs/tracker-functions#pageviews

Perhaps we should add it manually.

sidvishnoi (commented):
Two small details left. Rest is good!

cdn/src/lib/analytics.ts
kjmitchelljr: 👍🏿 change added: https://github.com/interledger/publisher-tools/pull/683/commits/61a6a32037b061549e4105b006d47bd62ed9a504

api/src/routes/events.ts
kjmitchelljr: Added in this commit: https://github.com/interledger/publisher-tools/pull/683/commits/61a6a32037b061549e4105b006d47bd62ed9a504

cdn/src/lib/analytics.ts
sidvishnoi: Let's move `screen` into event body (so it recomputes on each event), as browser window might be resized.

sidvishnoi:
I'll test it locally tomorrow, if works ok, will merge.

cdn/src/lib/analytics.ts
kjmitchelljr: [2dd0689](https://github.com/interledger/publisher-tools/pull/683/commits/2dd068962b2655a1b26c22319f3cb1f36cbd05aa) ✅

cdn/src/lib/analytics.ts
```diff
@@ -0,0 +1,23 @@
...
+    const body: EventBody = {
+      type: 'event',
+      payload: {
+        name: `embed.${tool}.${name}`,
+        url: `/embed/${tool}`,
+        screen: `${window.screen.width}x${window.screen.height}`,
+        data: { hostname, ...data },
+      },
+    }
+    navigator.sendBeacon?.(`${API_URL}/events`, JSON.stringify(body))
```
sidvishnoi: When sendBeacon isn't available, we should probably use regular fetch.

sidvishnoi (approved):
:shipit:

PR #684, opened 2026-04-24
fix(widget): pass `limits.receiver` to payment grant request

sidvishnoi:
As discussed in Open Payments call yesterday.


DarianM (approved):
read the OP call notes. :shipit:

DarianM:
what will be the default if incoming payment ID not specified?

sidvishnoi:
> what will be the default if incoming payment ID not specified?

Validation error, as required field.

DarianM:
tested that the outgoing interaction grant continues to work in dev and staging and `receiver` is passed correctly

PR #691, opened 2026-04-29
refactor(widget): separate API from widget logic using controller

sidvishnoi:
Follow up of https://github.com/interledger/publisher-tools/pull/694.


components/src/widget/views/interaction/interaction.ts
```diff
@@ -54,30 +43,63 @@ export class PaymentInteraction extends LitElement {
...
+    console.log(this.#controller?.getStatus)
+    try {
+      for await (const status of this.#controller.getStatus(
+        paymentId,
+        this.#pollingAbortController?.signal,
+      )) {
+        if (status.type === 'OUTGOING_PAYMENT_CREATED') {
+          this.currentView = 'processing'
+        } else if (status.type === 'OUTGOING_PAYMENT_DONE') {
+          this.currentView = 'success'
```
sidvishnoi: TODO: handle success/failure `status.result`

sidvishnoi: Done in https://github.com/interledger/publisher-tools/pull/691/commits/17dafc0cb8b3348c5e7ca0989a233ac2af8e5b3e.

components/src/widget/views/interaction/interaction.ts
```diff
@@ -54,30 +43,68 @@ export class PaymentInteraction extends LitElement {
...
+        if (status.type === 'OUTGOING_PAYMENT_CREATED') {
+          this.currentView = 'processing'
+        } else if (status.type === 'OUTGOING_PAYMENT_DONE') {
+          if (status.result === 'success') {
+            this.currentView = 'success'
+          } else {
+            this.currentView = 'failed'
+            this.errorMessage = status.error?.message || 'Payment failed'
+          }
+        } else if (status.type === 'PENDING_GRANT_INTERACTION') {
```
DarianM: let's remove this check of return early. `currentView`  default is 'authorizing'. makes the loop a bit easier to follow

sidvishnoi: I think it's safer to be explicit here, as we're exhausting all the potential `status.type` and handling each, instead of relying on defaults. But can move this check to bottom perhaps.

frontend/app/components/widget/WidgetPreview.tsx
```diff
@@ -39,11 +40,53 @@ export const WidgetPreview = ({
     }
   }, [profile, serviceUrls, opWallet])
 
+  useEffect(() => {
```
DarianM, line 44: Why were both this and `NO_OP_CONTROLLER` defined? maybe keep only one ?

sidvishnoi: `NO_OP_CONTROLLER` is meant used only as a sentinel value, when a controller isn't passed (so code doesn't crash). The WidgetPreview does add some more behavior than the `NO_OP_CONTROLLER`.
Perhaps we can reuse some of the `NO_OP_CONTROLLER` in this widget preview controller..

DarianM: I understand that `NO_OP_CONTROLLER` is sentinel, i didn't expect crash without defining some values. it's ok.

sidvishnoi: https://github.com/interledger/publisher-tools/pull/691/commits/a97c03cc1f961ec3194a2a0678443a1ad17f1b21

DarianM (approved):

PR #694, opened 2026-05-01
feat(widget): use new API routes, cleaner flow

sidvishnoi:
Use new API routes from https://github.com/interledger/publisher-tools/pull/692.

- There's no `finalize` now - the payment is triggered without client calling another endpoint, automatically immediately after grant redirect.
- Removed `postMessage` based payment status, using only the status from API.
   -  That endpoint doesn't do "long polling" - so we've to poll it repeatedly from client (`_pollPaymentCompletion` in `interaction.ts`).

Follow up:
- Make the entire flow cleaner. It was painful to make these changes - everything tied too deeply together and state flow not clear. Will use more components, improve using TypeScript, improve overall state flow - for devs as well as user.
- Remove unused files.


api/src/routes/payment/status2.ts
```diff
@@ -8,7 +8,7 @@ import { getData, setData, type PaymentKvData } from '../../utils/payments-kv'
 import { createHTTPException } from '../../utils/utils'
 
 app.get(
-  '/payment/status/:paymentId',
+  '/payment/status2/:paymentId',
```
kjmitchelljr, line 11: Is this URL endpoint temporary or a follow-up item?

frontend/app/utils/validate.server.ts
```diff
@@ -2,7 +2,7 @@ import type {
   PaymentStatus,
   PaymentStatusRejected,
   PaymentStatusSuccess,
-} from 'publisher-tools-api'
+} from 'publisher-tools-api/src/routes/payment/status'
```
kjmitchelljr, line 5: I assume the `../status2.ts` file will replace `status` or does this need be updated?

components/src/widget/views/confirmation/confirmation.ts
```diff
@@ -282,20 +277,26 @@ export class PaymentConfirmation extends LitElement {
...
-        receiveAmount: quote.receiveAmount,
+        receiver,
+        amount,
+        note = '',
+      } = this.configController.state
+      const { grantRedirectUrl, paymentId } = await this.initiatePayment({
+        sender: walletAddress,
+        receiver,
+        amount,
+        note,
```
kjmitchelljr: Is the note currently propagating through? This may be needed to make sure we capture it from the user input unless there is a state change elsewhere
```suggestion
note: this.note
```

components/src/widget/views/interaction/interaction.ts
```diff
@@ -134,86 +97,62 @@ export class PaymentInteraction extends LitElement {
...
+        this.requestUpdate()
+        await sleep(1500)
+        continue
+      }
+      if (res.type === 'OUTGOING_PAYMENT_DONE') {
+        this.currentView = 'success'
+        this.requestUpdate()
+        this._markPollingCompleted()
         return
       }
```
kjmitchelljr, line 130: For this `OUTGOING_PAYMENT_DONE` block should we add the `res.result` check. Basing on what I see in `api/src/routes/payment/status2.ts`
```js
{
  type: 'OUTGOING_PAYMENT_DONE'
  outgoingPaymentId: string
  result: 'success' | 'failure'
  error?: { code: string; message: string }
}
```

api/src/routes/payment/status2.ts
sidvishnoi: Yes, temporary endpoint. Accidentally used wrong route name in https://github.com/interledger/publisher-tools/pull/692. Next PR rename it back to status (as we remove existing status endpoint)

frontend/app/utils/validate.server.ts
sidvishnoi: Yes. `status2` will replace status, and most of this file will go away in https://github.com/interledger/publisher-tools/pull/695.

components/src/widget/views/confirmation/confirmation.ts
sidvishnoi: Good catch. We're setting it at line 293, but that's after use.

components/src/widget/views/interaction/interaction.ts
sidvishnoi: Done in https://github.com/interledger/publisher-tools/pull/694/commits/e39ea379403a76aed81aa4a07be32f7b59938d06

components/src/widget/views/confirmation/confirmation.ts
sidvishnoi: Done in https://github.com/interledger/publisher-tools/pull/694/commits/fc8de098c9a0b8f7cb39c361fc5182f626daabec

components/src/widget/views/interaction/interaction.ts
DarianM: I might have created this status 😓 but one thing I do not like about this is the type assumes if success type can happen then it can also have an `error`, and do we want the `outgoingPaymentId` with result failure?
so a cleaner version might be: 
```
  | { type: 'OUTGOING_PAYMENT_SUCCESS'; outgoingPaymentId: string }
  | { type: 'OUTGOING_PAYMENT_FAILURE'; error: { code: string; message: string } }
```
treat this optional comment

sidvishnoi: We already have `outgoingPaymentId` from a previous status, but can add here as well - there's no user of that `outgoingPaymentId` as of yet though. So, will add when we need it to `OUTGOING_PAYMENT_DONE`.

DarianM (approved):

PR #695, opened 2026-05-01
chore: remove now-unused code due to new payment API

sidvishnoi:
Follow-up of https://github.com/interledger/publisher-tools/pull/691 and https://github.com/interledger/publisher-tools/pull/694.


api/src/routes/payment/status2.ts
DarianM, line 1: If we rename this file back now, it will conflict with the name of the deleted file, right?

sidvishnoi: Yes (https://github.com/interledger/publisher-tools/pull/695/commits/190f4c8a84c47e624a58a54c20501ad44ab986be).

DarianM (approved):

PR #700, opened 2026-05-05
refactor(components/widget): extract home view component

sidvishnoi:
Extracted from https://github.com/interledger/publisher-tools/pull/696


components/src/widget/widget.css
```diff
@@ -190,85 +190,6 @@ a:focus-visible {
...
-  display: flex;
-  flex-direction: column;
-  width: 100%;
-  box-sizing: border-box;
-  padding: 0px var(--Spacings-2xl);
-  flex: 1 1 auto;
-  gap: var(--Spacings-lg);
-}
-
 .widget-body p {
```
DarianM: we can move this `.widget-body p` too  ->`home.css`

components/src/widget/views/home/home.ts
```diff
@@ -0,0 +1,137 @@
...
+  @property({ type: String }) ctaText = ''
+  @property({ type: String }) description = DEFAULT_DESCRIPTION
+  @property({ type: Boolean }) showDescription = true
+  @property({ type: String }) backgroundColor = '#fff'
+
+  @state() private _walletAddress = ''
+  @state() private _error = ''
+  @state() private _isSubmitting = false
+
+  static styles = [unsafeCSS(stylesBase), unsafeCSS(styles)]
```
DarianM:
```suggestion
  static styles = [unsafeCSS(styles)]
```
we can drop the stylesBase

components/src/widget/views/home/home.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  render() {
+    const hasError = !!this._error
+
+    const descriptionElement = this.showDescription
+      ? html`<p>${this.description ?? DEFAULT_DESCRIPTION}</p>`
+      : html`<div class="divider" />`
+
+    return html`
+      <div class="widget-header-container">
```
DarianM, line 81: in the old code before extracting, `.widget-header-container` and `.payment-form.widget-body` were direct flex children of container div `.content`.
Now there's <wm-payment-home> as an intermediary element with its own shadow DOM boundary. some styling is affected with this extract.
something like
```css
:host {
  display: flex;
  flex-direction: column;
  flex: 1;
  min-height: 0;
}
``` 
to home.css is missing to have the old design back

DarianM (changes requested):

components/src/widget/views/home/home.ts
```diff
@@ -0,0 +1,137 @@
...
+            ?disabled=${this._isSubmitting}
+            aria-invalid=${hasError}
+            aria-describedby=${hasError ? 'wallet-error' : ''}
+          />
+
+          <div
+            id="wallet-error"
+            class="error-message"
+            role="alert"
+            aria-live="polite"
```
kjmitchelljr: Would the `aria-live="polite"` still be needed if we have `role="alert"` or is it meant to override in this context?

components/src/widget/views/home/home.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  private onInput() {
+    this._error = ''
+  }
+
+  render() {
+    const hasError = !!this._error
+
+    const descriptionElement = this.showDescription
+      ? html`<p>${this.description ?? DEFAULT_DESCRIPTION}</p>`
```
kjmitchelljr: Do we want this to be `| |` similar to `this.title` otherwise since it's being set to the default towards the top of the file we probably wouldn't need the fallback

sidvishnoi: We do need the fallback, for example, in case of `title`, if we pass `.title=""` (during preview), we still want to show the default title instead of empty. The defaults only work if if prop isn't passed at all.

components/src/widget/views/home/home.ts
sidvishnoi: Good catch. I earlier added `role="status"`, then changed it to alert, but missed `aria-live`. With `role="alert"`, `aria-live` means `"assertive"`.

components/src/widget/views/home/home.ts
sidvishnoi: I was trying hard to notice the difference; turns out, the "Support me" button had moved above without this. Thanks for catching this!

DarianM: apologies for ur time. i'll post before/after screens next time 👍

DarianM (approved):

Issue #672, opened 2026-04-20
Integrate OTP for paywalls into a publisher's website


RabebOthmani:
As a publisher, I want to integrate the one-time payment tool into my website, so that I can start monetizing paywalled content

Issue #670, opened 2026-04-20
[Feature] One Time Payments for Paywalls


RabebOthmani:
Paywalls are one of the most common and established ways publishers monetize their content online. In keeping with our approach of solving real publisher problems and being present where there is genuine opportunity, 
we are therefore introducing a new Publisher Tool that allows visitors to make a one-time payment to unlock content hidden behind a paywall, powered by Open Payment.

PR #702, opened 2026-05-06
feat(paywall): add base component, API interface, script

### LINKED ISSUE #677: Unlock content for users who already paid for it

Image

sidvishnoi:
Part of https://github.com/interledger/publisher-tools/issues/670
Part of https://github.com/interledger/publisher-tools/issues/672
Part of https://github.com/interledger/publisher-tools/issues/677


cdn/src/paywall.ts
```diff
@@ -0,0 +1,104 @@
...
+      const redirectUrl = window.location.href
+      return initiatePayment(API_URL, {
+        sender,
+        receiver,
+        receiveAmount,
+        note,
+        redirectUrl,
+      })
+    },
+    async *getStatus(paymentId, signal) {
```
DarianM, line 53: duplicate of cdn `widget.ts` .
`getStatus` is identical in both files. since `initiatePayment`, `fetchQuote`, and `getWallet` are already extracted to a shared module, this should follow the same pattern maybe ?

sidvishnoi: Was thinking if we'll need this in same manner as widget. If we need it as same as widget, will extract then. So leaving as is until we actually implement it.

DarianM (approved):

PR #703, opened 2026-05-06
fix(api): correctly format ZodError responses

sidvishnoi:
```diff
# GET /profile/unknown-tool
--- text	Original
+++ text	Modified
@@ -1,7 +1,15 @@
 {
-  "success": false,
   "error": {
-    "name": "ZodError",
-    "message": "[\n  {\n    \"code\": \"invalid_value\",\n    \"values\": [\n      \"banner\",\n      \"widget\",\n      \"offerwall\"\n    ],\n    \"path\": [\n      \"tool\"\n    ],\n    \"message\": \"Invalid option: expected one of \\\"banner\\\"|\\\"widget\\\"|\\\"offerwall\\\"\"\n  }\n]"
+    "message": "Validation failed",
+    "code": "VALIDATION_ERROR",
+    "details": {
+      "issues": [
+        {
+          "path": "tool",
+          "message": "Invalid option: expected one of \"banner\"|\"widget\"|\"offerwall\"",
+          "code": "invalid_value"
+        }
+      ]
+    }
   }
 }
```

### DIFF
--- a/api/src/app.ts
+++ b/api/src/app.ts
@@ -21,7 +21,8 @@ app.use('*', async (c, next) => {
   await next()
   if (c.res.status !== 400) return
 
-  const data = await c.res.clone().json()
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const data: any = await c.res.clone().json()
   if (data?.error?.name === 'ZodError' && data.error.message?.startsWith('[')) {
     c.res = c.json(
       {
--- a/api/src/app.ts
+++ b/api/src/app.ts
@@ -17,6 +17,22 @@ export type Env = {
 
 export const app = new Hono<{ Bindings: Env }>()
 
+app.use('*', async (c, next) => {
+  await next()
+  if (c.res.status !== 400) return
+
+  const data = await c.res.clone().json()
+  if (data?.error?.name === 'ZodError' && data.error.message?.startsWith('[')) {
+    c.res = c.json(
+      {
+        ...data,
+        error: { ...data.error, message: JSON.parse(data.error.message) },
+      },
+      c.res.status,
+    )
+  }
+})
+
 app.use(
   '*',
   cors({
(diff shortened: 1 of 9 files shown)


api/src/app.ts
DarianM, line 44: hmm. since we need the above middleware instead to catch the zod validation erros this `if` block (line 61) is effectively dead code.
I do not like the middleware as  is prone to something going wrong moving forward if zod decides to change something. I still like this `err` object you created for `error instanceof ZodError`.
hear me out: https://hono.dev/docs/guides/validation#zod-validator-middleware
what if instead of the `zValidator` we use for api routes we build a custom `validate` wrapper, it accepts a hook as a third argument specifically for this. 
```ts
export const validate = <T extends ZodSchema>(
  target: 'param' | 'query' | 'json',
  schema: T,
) =>
  zValidator(target, schema, (result, c) => {
    if (!result.success) {
      return c.json(
        {
          error: {
            message: 'Validation failed',
            code: 'VALIDATION_ERROR',
            details: {
              issues: result.error.issues.map((issue) => ({
                path: issue.path.join('.'),
                message: issue.message,
                code: issue.code,
              })),
            },
          },
        },
        400,
      )
    }
  })
```
then in routes, we'd need to swap `zValidator` -> `validate` . the output shape would then match this nice `err` obj format.

kjmitchelljr: I follow, this makes sense. Thank you for the docs @DarianM.

kjmitchelljr (commented):
Darian's suggestion seems to be aligned with what we're trying to resolve here

api/src/app.ts
sidvishnoi: I considered this, but avoided doing so as that meant making changes in all routes instead of one place.
Btw, if we throw from the third-argument, like:
```ts
zValidator(target, schema, (result) => {
  if (!result.success) throw result.error
}
```
we can handle it this as `error instanceof ZodError` in our existing middleware, instead of handling in validator.

> I do not like the middleware as is prone to something going wrong moving forward if zod decides to change something.

I agree. Let's go with the throw in custom validator, and handle error in middleware; instead of JSON.parse in middleware. Sounds good?

DarianM (approved):

PR #704, opened 2026-05-07
refactor(components/widget): extract `amount` component

sidvishnoi:
Extracted from https://github.com/interledger/publisher-tools/pull/696

Cleanup opportunity for future: create a base.css which includes some common styles that we use across components. Right now, I'm importing confirmation.css as some `.form-field` styles are used in this amount as well as "Note" input.

### DIFF
--- a/components/src/widget/views/amount/amount.css
+++ b/components/src/widget/views/amount/amount.css
@@ -1,0 +1,115 @@
+:host {
+  font-family: var(--wm-font-family, system-ui, sans-serif);
+  display: block;
+  flex: 0;
+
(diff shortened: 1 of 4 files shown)


DarianM (approved):
Tested. looks good. much improved. a small improvement nit. if I hit Delete and type the exact same amount again in the input field, it shouldn't fetch quote because I already have it for that amount.

sidvishnoi:
>  it shouldn't fetch quote because I already have it for that amount

Good catch. Will look into caching in general, not just on last amount, as a follow-up.

PR #718, opened 2026-05-14
feat(analytics): track settings changed event(s)

### LINKED ISSUE #693: Track changed profile fields on save (settings_changed event)
Restore the `settings_changed` event that was removed during PR #664 review, this time tracking *which* fields actually changed on save rather than firing on every save.

## Event shape

```ts
trackEvent('banner_settings_changed', {
  tool: 'banner',
  'field.font.name': true,        // boolean changes -> true
  'field.font.size': true,
  'field.title.text': 50,      
  'field.color.background': 7,
})
```

- Event name: `${tool}_settings_changed` for `banner`, `widget`, `offerwall` (matches the `${tool}_profile_saved` / `${tool}_script_generated` convention so it's filterable in the Umami dashboard).
- Only fire when there is at least one changed field.

## Scope

- [] Helper `diffProfile(prev, current)` that walks nested profile objects and produces a flat `ChangedFields` map. String leaves -> length, other primitives -> `true`. Skips `$`-prefixed keys.
- [] Extend `EventRegistry` in `frontend/app/lib/analytics.ts` with `${Tool}_settings_changed`.
- [] Each tool store (`banner-store`, `widget-store`, `offerwall-store`) exposes `actions.getChangedFields()` that diffs the active tab's saved snapshot against the current proxy state.
- [] `useSaveProfile` captures changed fields *before* `commitProfile()` and fires `${tool}_settings_changed` on successful save when there are actual changes.

## Acceptance criteria

- Saving a profile with no changes fires `${tool}_profile_saved` but does *not* fire `${tool}_settings_changed`.
- Changing `font.name` and `font.size` produces two distinct `field.font.name` / `field.font.size` keys (not a single `font: true`).
- Text fields report length (e.g. `field.title.text: 50`), not the raw value.
- Verified locally with `pnpm analytics:dev` that events arrive in Umami with the expected per-field properties.

## Files touched

- `frontend/app/utils/profile-diff.ts` (new)
- `frontend/app/lib/analytics.ts`
- `frontend/app/stores/{banner,widget,offerwall}-store.ts`
- `frontend/app/hooks/useSaveProfile.tsx`

kjmitchelljr:
## Summary
Implements detailed event tracking for settings changes across Banner, Offerwall, and Widget tools

### Changes
* **State Diffing:** Added a recursive utility to compare profile states, tracking modified fields
* **Centralized Store Logic:** Refactored `createToolStoreUtils` to integrate diffing directly into the profile commit
* **Analytic Decision:** Updated the save hook to trigger `${tool}_settings_changed` events only when changes occur

### Related Issues
Closes #693


frontend/app/utils/profile-diff.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// $version and $modifiedAt are system metadata; $name is user-editable (profile rename)
+const SKIPPED_KEYS = new Set(['$version', '$modifiedAt'])
+
+export function diffProfile(
+  prev: object | undefined,
+  current: object,
+): ChangedFields {
+  const result: Record<string, boolean | string> = {}
+  walk(prev as PlainObject | undefined, current as PlainObject, '', result)
```
sidvishnoi: Consider using this package: https://www.npmjs.com/package/microdiff

frontend/app/lib/analytics.ts
```diff
@@ -19,3 +24,12 @@ export function useTrackEvent(): TrackFn {
...
   }
 }
+
+export function trackSettingsEvent(
+  trackFn: TrackFn,
+  tool: Tool,
+  changed: ChangedFields,
+): void {
+  if (Object.keys(changed).length === 0) return
+  trackFn(`${tool}_settings_changed`, { changed })
```
sidvishnoi: Do we want this `changed.` prefix in all properties given event name says `_changed`?

frontend/app/lib/analytics.ts
```diff
@@ -19,3 +24,12 @@ export function useTrackEvent(): TrackFn {
     window.umami?.track(eventName, { tool, ...eventData })
   }
 }
+
+export function trackSettingsEvent(
```
sidvishnoi: Given this is used only in `useSaveProfile`, consider inlining it there instead of adding to general `analytics`.

frontend/app/lib/analytics.ts
```diff
@@ -8,6 +12,7 @@ export type EventRegistry =
   | [event: 'link_tag_generated', data: { wallet_provider: string }]
   | [event: `${Tool}_profile_saved`]
   | [event: `${Tool}_script_generated`]
+  | [event: `${Tool}_settings_changed`, data: SettingsChangedData]
```
sidvishnoi: Should we include this data in saved/generated events instead of a new event? Given either of these events will be fired immediately after this change event.

frontend/app/utils/utils.store.ts
```diff
@@ -145,6 +153,29 @@ export function createToolStoreUtils<T extends Tool>(
...
 
+    // User-initiated save; returns diff for analytics
+    commitActiveProfile(activeTab: ProfileId): ChangedFields {
+      const prev = snapshots.get(activeTab)
+      const current = snapshot(store.profiles[activeTab]) as ToolProfile<T>
+      const changed = diffProfile(prev, current)
+
+      snapshots.set(activeTab, current)
+      store.profilesUpdate.delete(activeTab)
+      persistSnapshots()
```
sidvishnoi, line 166: Why persist all snapshots when committing only active profile?

frontend/app/stores/widget-store.ts
```diff
@@ -116,23 +114,10 @@ export const actions = {
     )
   },
   commitProfile() {
-    const profile = snapshot(widget.profile)
-    snapshots.set(toolState.activeTab, profile)
-    widget.profilesUpdate.delete(toolState.activeTab)
-
-    const snaps = Object.fromEntries(snapshots.entries())
-    localStorage.setItem(snapshotsStorageKey, JSON.stringify(snaps))
+    return widgetStoreUtils.commitActiveProfile(toolState.activeTab)
```
sidvishnoi, line 117: Can you send these refactors as a separate PR? Then we can focus on analytics separately.

frontend/app/utils/profile-diff.ts
```diff
@@ -0,0 +1,47 @@
+export type ChangedFields = Partial<Record<`field.${string}`, boolean | string>>
```
sidvishnoi: This will take up less space (visually as well as in storage):
```suggestion
export type ChangedFields = Partial<Record<`f.${string}`, boolean | string>>
```

sidvishnoi (changes requested):

frontend/app/stores/widget-store.ts
kjmitchelljr: https://github.com/interledger/publisher-tools/pull/734/changes added in this PR

frontend/app/utils/utils.store.ts
kjmitchelljr: @sidvishnoi I can verify why this is the case, initially I just took the logic that was in `commitProfile()` in the separate stores and brought it over. Since the `localStorage.setItem()` was being called twice once in `commitProfile()` and once `commitAllProfiles()` I decided to wrap in a function, but could leave as it is only a one-liner

sidvishnoi: Something to address later on..

frontend/app/lib/analytics.ts
kjmitchelljr: Yep, that makes sense to me since it isn't firing as user makes the changed

frontend/app/utils/profile-diff.ts
kjmitchelljr: ✅ [9ff6430](https://github.com/interledger/publisher-tools/pull/718/commits/9ff643064eab9da2da36350fafe13c64c9c572c0) added here

frontend/app/utils/utils.store.ts
kjmitchelljr: [9ff6430](https://github.com/interledger/publisher-tools/pull/718/commits/9ff643064eab9da2da36350fafe13c64c9c572c0) added TODO here

frontend/app/lib/analytics.ts
kjmitchelljr: ✅ - [f6ce79c](https://github.com/interledger/publisher-tools/pull/718/commits/f6ce79c12faaf9c05059eb4e79ee4207fbb037ea)

frontend/app/utils/profile-diff.ts
kjmitchelljr: ✅ - [f6ce79c](https://github.com/interledger/publisher-tools/pull/718/commits/f6ce79c12faaf9c05059eb4e79ee4207fbb037ea)

frontend/app/lib/analytics.ts
kjmitchelljr: ✅ - [f6ce79c](https://github.com/interledger/publisher-tools/pull/718/commits/f6ce79c12faaf9c05059eb4e79ee4207fbb037ea)

frontend/app/utils/profile-diff.ts
```diff
@@ -0,0 +1,27 @@
...
+// $version and $modifiedAt change on every save
+const SKIPPED_KEYS = new Set(['$version', '$modifiedAt'])
+
+export function diffProfile(prev: object, current: object): ChangedFields {
+  const result: Record<string, boolean | string> = {}
+  for (const d of diff(prev, current)) {
+    if (d.type !== 'CHANGE') continue
+    if (SKIPPED_KEYS.has(d.path[0] as string)) continue
+    result[`f.${d.path.join('.')}`] =
+      typeof d.value === 'string' ? lengthBucket(d.value) : true
```
sidvishnoi: I think we might want some details specific to our config here. For example,
- with animation, even though a string, having its length isn't meaningful. This applies to all "enum" types. We can show the value used directly, as not user specific.
- With title options (when not provided by user), if we use full text (or some ID), we know which of the title options get used by users most often.
- With booleans, knowing if set to true/false would be helpful.
- What about numbers? For example, in paywall, we'll have a currency (value) - it's a numeric string.

The privacy concern is with user-provided text mostly. Also, there, the values can be arbitrary - so we won't get insights without some sort of bucketing.

frontend/app/utils/profile-diff.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+  return result as ChangedFields
+}
+
+function lengthBucket(s: string): string {
```
sidvishnoi: Do we want custom bucket sizes for title and description?

frontend/app/lib/analytics.ts
```diff
@@ -1,13 +1,14 @@
...
 
 export type EventRegistry =
   | [event: 'click_card_tool', data: { link: string }]
   | [event: 'wallet_connected', data: { wallet_provider: string }]
   | [event: 'wallet_disconnected']
   | [event: 'link_tag_generated', data: { wallet_provider: string }]
-  | [event: `${Tool}_profile_saved`]
-  | [event: `${Tool}_script_generated`]
+  | [event: `${Tool}_profile_saved`, data: { changed: ChangedFields }]
+  | [event: `${Tool}_script_generated`, data: { changed: ChangedFields }]
```
sidvishnoi: In Umami, this will add `changed.` prefix for each files. e.g. `changed.f.title.text`. Do we want that or, `f.title.text` is enough as data, with us knowing that this corresponds to "change"?

sidvishnoi (changes requested):

pnpm-lock.yaml
sidvishnoi, line 1: Can you undo the `pnpm dedupe`? For this PR, I think there should only be microdiff related changes to this file.

sidvishnoi (commented):
I'll test locally, and check how does it work with multiple profiles being edited, but this looks great!

sidvishnoi (approved):
Works well.
Given we save only active profile (not all profiles), our events work well.

PR #720, opened 2026-05-19
feat(paywall): add D1 database for storing payments

### LINKED ISSUE #677: Unlock content for users who already paid for it

sidvishnoi:
Part of https://github.com/interledger/publisher-tools/issues/677

### DIFF
--- a/api/src/utils/payments-db.ts
+++ b/api/src/utils/payments-db.ts
@@ -139,7 +139,7 @@ export async function UNSAFE_devEmptyDatabase(db: D1Database) {
 }
 
 function mapStatusToId(status: PaymentStatus) {
-  return status.toLowerCase() === 'created' ? 0 : 1
+  return status === 'complete' ? 1 : 0
 }
 
 function mapStatusIdToStatus(status: 0 | 1): PaymentStatus {
--- a/api/src/utils/payments-db.ts
+++ b/api/src/utils/payments-db.ts
@@ -2,7 +2,8 @@ import type { D1Database } from '@cloudflare/workers-types'
 import type { Amount } from '@shared/types'
 import type { WalletAddressInfo } from '../types'
 
-const sql = String.raw // for syntax highlighting
+/** for syntax highlighting */
+export const sql = String.raw
 
 export async function savePayment(db: D1Database, data: Payment) {
   const site = getSite(data.url)
--- a/api/src/utils/payments-db.ts
+++ b/api/src/utils/payments-db.ts
@@ -52,24 +52,35 @@ export async function setPaymentStatus(
   db: D1Database,
   paymentId: Payment['paymentId'],
   status: PaymentStatus | 'failed',
-) {
+): Promise<boolean> {
   if (status === 'failed') {
     // If failed (money stayed in user wallet), delete from both tables.
-    await db.batch([
+    // The first call deletes from meta table only if `status = 0` in the main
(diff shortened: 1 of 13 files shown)


DarianM (approved):

api/test/apply-migrations.ts
```diff
@@ -0,0 +1,9 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck doesn't work for some reason
```
kjmitchelljr, line 2: Read it may have to do with ordering
```suggestion
// @ts-nocheck doesn't work for some reason
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```

api/src/utils/payments-db.ts
```diff
@@ -0,0 +1,204 @@
...
+export async function UNSAFE_devEmptyDatabase(db: D1Database) {
+  await db.batch([
+    db.prepare(`DELETE FROM paywall_payments_meta`),
+    db.prepare(`DELETE FROM paywall_payments`),
+  ])
+}
+
+function mapStatusToId(status: PaymentStatus) {
+  return status.toLowerCase() === 'created' ? 0 : 1
+}
```
kjmitchelljr, line 154: If value is something else (not `created`) what does the the 0 or 1 setting represent, e.g. does 1 assume `complete`?

api/src/utils/payments-db.ts
```diff
@@ -0,0 +1,204 @@
...
+    // If failed (money stayed in user wallet), delete from both tables.
+    await db.batch([
+      db
+        .prepare(`DELETE FROM paywall_payments WHERE paymentId = ?`)
+        .bind(paymentId),
+      db
+        .prepare(`DELETE FROM paywall_payments_meta WHERE paymentId = ?`)
+        .bind(paymentId),
+    ])
+    return
```
kjmitchelljr: nit: perhaps also check `status = 0`

kjmitchelljr (commented):
Small things - but could merge and follow up if anything sticks out.

api/src/utils/payments-db.ts
sidvishnoi: Yes, it's either 0 or 1 (as in Database). If not created, it's complete. But I'll update it other way around - so we don't accidentally mark as complete if something else is passed.

api/src/utils/payments-db.ts
sidvishnoi: Good idea! Don't want to accidentally delete already complete one due to some timing issue.

api/test/apply-migrations.ts
sidvishnoi: The eslint rule is to tell eslint typescript, that the _following_ ts-nocheck is ok.

api/src/utils/payments-db.ts
sidvishnoi: https://github.com/interledger/publisher-tools/pull/720/commits/9138c162cbb3205fed750fb55ed3a38363a247c6

api/src/utils/payments-db.ts
sidvishnoi: https://github.com/interledger/publisher-tools/pull/720/commits/c7fe452cb3201774cfe6e7daf3b4e61526e3ccdc

PR #723, opened 2026-05-20
feat(paywall): store payments in db for future entitlement

### LINKED ISSUE #677: Unlock content for users who already paid for it

sidvishnoi:
Major part of https://github.com/interledger/publisher-tools/issues/677

- The `/paywall/redirect` API endpoint, which is passed to `/payment/initiate` handles the payment storage to our relational database (introduced in [#720](https://github.com/interledger/publisher-tools/pull/720), using additional data from [#722](https://github.com/interledger/publisher-tools/pull/722))
   - It eventually sends us back to original page URL via the `next` query param.
- The `/paywall/status` endpoint is mostly same as `/payment/status` endpoint in terms of client-side behavior (polling for complete).
   - But it can additionally get payment from the relational db, to check for payment completion (the incoming payment (not outgoing, which requires us storing sender wallet address in plain text)) even after payment details are deleted from temporary-KV.
   - This endpoint is also responsible for marking payment as "complete" in DB.


api/src/routes/paywall/redirect.ts
sidvishnoi, line 1: This endpoint ensures the payment is saved behind-the-scenes after grant accept, without needing a call from client.

api/src/routes/paywall/redirect.ts
```diff
@@ -0,0 +1,56 @@
...
+      outgoingPaymentId,
+      sender,
+      receiver,
+      status: data.status,
+      amount: data.amount,
+    })
+  } catch (err) {
+    console.error(err)
+    return redirect(
+      urlWithParams(next, { paymentId, status: 'failure', reason: 'internal' }),
```
kjmitchelljr:
```suggestion
      urlWithParams(next, { paymentId, result: 'failure', reason: 'internal' }),
```

api/src/routes/paywall/redirect.ts
```diff
@@ -0,0 +1,56 @@
...
+import { urlWithParams } from '@shared/utils'
+import { app } from '../../app'
+import { PaymentIdSchema } from '../../schemas/payment'
+import { savePayment } from '../../utils/payments-db'
+import { getData } from '../../utils/payments-kv'
+import { validate } from '../../utils/utils'
+
+const schema = z.object({
+  // Passed by client
+  next: z.url(),
```
kjmitchelljr, line 11: Can you explain the `next()` variable a bit more? Any security issue we would have to consider if coming in as query parameter?

cdn/src/utils/paywall-utils.ts
```diff
@@ -0,0 +1,7 @@
+import { ensureEnd } from '@shared/utils'
+
+export function getPageUrl(url: URL | Location) {
+  const { origin, pathname } = url
+  const path = /\.html?^/.test(pathname) ? pathname : ensureEnd(pathname, '/')
```
kjmitchelljr: Right now this appears to append a `/` if the path ends in `.html` or `.htm` resulting is something like https://example.com/x.html -> https://example.com/x.html/ - is this the expected behavior?

api/src/routes/paywall/status.ts
```diff
@@ -0,0 +1,164 @@
...
+    const openPayments = await OpenPaymentsService.getInstance(env)
+    const result = await openPayments.pollOutgoingPayment(
+      data.outgoingPaymentId,
+      data.outgoingPaymentGrantAccessToken,
+    )
+    await setData(
+      env.PUBLISHER_TOOLS_KV,
+      paymentId,
+      {
+        status: 'COMPLETE',
```
kjmitchelljr: Could prematurely set to `COMPLETE` even if failed which would cause an issue on line 82. Perhaps checking for result earlier and returning the result state in the CREATED branch

api/src/routes/paywall/redirect.ts
sidvishnoi: This is sent as part of initiatePayment essentially. We've to send user to the page they came from. 
Security risk: someone will have to create an API call to our endpoint with some other next param, then make user click that grant URL so they can be redirected to some other URL. But then the payment will be marked for that next URL only - so the url that's passed as next will be the one we mark paid as.

cdn/src/utils/paywall-utils.ts
sidvishnoi: Should've been $ in regex, not ^

api/src/routes/paywall/status.ts
sidvishnoi: https://github.com/interledger/publisher-tools/pull/723/commits/546514b683e19fc774f5af2c5c029a7cf139502f

api/src/routes/paywall/redirect.ts
sidvishnoi: https://github.com/interledger/publisher-tools/pull/723/commits/0916bd338e9fcfc5bb792a747c53675d71a86c62

cdn/src/utils/paywall-utils.ts
sidvishnoi: https://github.com/interledger/publisher-tools/pull/723/commits/ce4965330cad969b9f18f8b9e8cd745c245be4b5

DarianM (approved):

PR #727, opened 2026-05-25
feat(paywall): integrate backend API with embed

### LINKED ISSUE #672: Integrate OTP for paywalls into a publisher's website

### LINKED ISSUE #677: Unlock content for users who already paid for it

sidvishnoi:
Closes https://github.com/interledger/publisher-tools/issues/672
Closes https://github.com/interledger/publisher-tools/issues/677

Preview: https://test.sidvishnoi.com/web-monetization/publisher-tools/paywall.html?src=727

Follow up: error handling, payment rejected etc.

### DIFF
--- a/cdn/src/paywall.ts
+++ b/cdn/src/paywall.ts
@@ -65,6 +65,12 @@ function main() {
       })
       const data: PaywallEntitlementResult = await res.json()
       if (data.token) storage.authJwt.set(data.token)
+      if (data.entitlement === 'pending') {
+        actions.setView('verify', {
+          paymentId: data.paymentId,
+          sender: walletAddress!,
+        })
+      }
 
       return data.entitlement
     },
--- a/cdn/src/paywall.ts
+++ b/cdn/src/paywall.ts
@@ -139,8 +139,8 @@ function main() {
     },
   })
 
-  document.body.appendChild(element)
   handlePageLoad(pageUrl, paramsFromUrl, actions)
+  document.body.appendChild(element)
 }
 
 function handlePageUrlOnLoad() {
--- a/cdn/src/paywall.ts
+++ b/cdn/src/paywall.ts
@@ -51,7 +51,7 @@ function main() {
       console.debug('checkEntitlement', { walletAddress })
       const token = storage.authJwt.get()
       if (!walletAddress && !token) {
-        return 'no-access'
+        return { entitlement: 'no-access' }
       }
       const url = urlWithParams(new URL('/paywall/entitlement', API_URL), {
         url: pageUrl,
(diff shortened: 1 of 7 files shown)


cdn/src/utils/index.ts
```diff
@@ -144,6 +144,54 @@ export function appendPaymentPointer(walletAddressUrl: string) {
...
+  }
+  if (typeof payload === 'object' && payload !== null) {
+    return payload as T
+  }
+  return null
+}
+
+export function redirect(url: string): never {
+  window.location.href = url
+  throw 'unreachable'
```
kjmitchelljr, line 192: Don't think this currently affects the overall flow, but perhaps changing to return a Promise since `authenticate()` is async, e.g. 
```suggestion
return new Promise<never>(() => {})
```

cdn/src/paywall.ts
```diff
@@ -27,15 +42,48 @@ function drawPaywall() {
...
-  element.setController({
+  const actions = element.setController({
     receiverWalletAddressUrl: params.walletAddress,
+    senderWalletAddressUrl: storage.authJwt.getWalletAddress(),
     cdnUrl: params.cdnUrl,
     fetchConfig: () => fetchProfile(API_URL, 'paywall', params),
-    async checkEntitlement() {
-      return 'no-access' // TODO: create and call API
+    async checkEntitlement(walletAddress) {
+      console.debug('checkEntitlement', { walletAddress })
```
kjmitchelljr, line 51: Note to remove before pushing to prod - I assume needed for initial testing?

cdn/src/paywall.ts
```diff
@@ -91,5 +144,117 @@ function drawPaywall() {
...
+    // Set when result=failure typically
+    res.reason = params.reason
+    url.searchParams.delete('reason')
+  }
+
+  if (window.location.href !== url.href) {
+    window.history.replaceState(null, '', url.href)
+  }
+
+  console.debug('handlePageUrlOnLoad', res)
```
kjmitchelljr, line 180: Note to remove before pushing to prod - I assume needed for initial testing?

components/src/paywall/index.ts
```diff
@@ -126,15 +162,37 @@ export class Paywall extends LitElement {
...
-    if (!this.#controller.isPreviewMode) {
-      window.location.href = result.grantRedirectUrl
+  async #onPaymentConfirmed(
+    ev: CustomEvent<PaymentVerifyEvents['payment_confirmed']>,
+  ) {
+    const sender = ev.detail.sender
+    const status = await this.#controller.checkEntitlement(sender)
+    if (status.entitlement === 'has-access') {
+      this.remove()
+      return
```
kjmitchelljr, line 194: In the view would there be checks added for the other use cases: `auth-required`, `pending`, and `no-access`?

cdn/src/paywall.ts
sidvishnoi: Prod is far away, but yes, these will be removed once tested enough, including in staging.

components/src/paywall/index.ts
sidvishnoi: Those cases will be impossible, but in case it does and for other cases, I'll add error logging/screens as follow-up.

cdn/src/utils/index.ts
sidvishnoi: Not really needed, as this part will never be reached. 
`authenticate()` being async doesn't matter here - this is a function in its own, and its not returned from `authenticate()`

kjmitchelljr (approved):
👍🏿

PR #746, opened 2026-06-02
feat(app/paywall): show PaywallPreview in editor

### LINKED ISSUE #690: Show live preview
As a publisher, I want to be able to see a live preview of the paywall  while I'm customizing it
So I know how it would look like for visitors 

Acceptance Criteria:

- Preview should reflect changes to the paywall live
- Preview should reflect customization both to the settings (price, screen ) and look of the paywall
- Preview article is a dummy article for illustration purposes and not the real content from the publisher

UI

Image

sidvishnoi:
Closes https://github.com/interledger/publisher-tools/issues/690

Add a new route "paywall preview", which gets embedded as `<iframe>` in PaywallPreview. Adding as iframe lets us use the media queries/responsive styles as if the paywall was embedded on a website. It also lets us avoid any "preview mode" styles added to our components. `PaywallPreview` passes the `profile` to the iframe via `postMessage` API.

### DIFF
--- a/frontend/app/routes/paywall-preview.tsx
+++ b/frontend/app/routes/paywall-preview.tsx
@@ -1,0 +1,74 @@
+import { useEffect, useState } from 'react'
+import { NO_OP_CONTROLLER } from '@c/paywall/controller'
+import { getDefaultProfile } from '@shared/default-data'
+import { CDN_URL } from '@shared/defines'
+import type { Message } from '~/components/paywall/PaywallPreview'
+
+export default function PaywallPreview() {
+  const [profile, setProfile] = useState(() => getDefaultProfile('paywall'))
+  const [isLoaded, setIsLoaded] = useState(false)
+
+  useEffect(() => {
+    document.querySelector('body > header')?.remove()
+    document.querySelector('body > footer')?.remove()
+  }, [])
+
+  const NAME = 'wm-paywall'
+  useEffect(() => {
+    const load = async () => {
+      if (!customElements.get(NAME)) {
+        const { Paywall } = await import('@tools/components/paywall/index')
+        if (!customElements.get(NAME)) {
+          customElements.define(NAME, Paywall)
+        }
+      }
+      setIsLoaded(true)
+    }
+    load()
+  }, [])
+
+  useEffect(() => {
+    if (!isLoaded) return
+
+    const el = document.createElement(NAME)
+    const _actions = el.setController({
+      isPreviewMode: true,
+      authenticate: NO_OP_CONTROLLER.authenticate,
+      cdnUrl: CDN_URL,
+      checkEntitlement: NO_OP_CONTROLLER.checkEntitlement,
+      fetchConfig: () => Promise.resolve(profile),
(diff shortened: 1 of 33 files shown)


frontend/app/routes/paywall.tsx
sidvishnoi, line 1: Changes in this file move PaywallPreview to be at start (near Configure builder), instead of starting in parallel to "content/appearance builder".

sidvishnoi: Added some `<>` to avoid too many whitespace changes.

frontend/app/utils/utils.storage.ts
```diff
@@ -64,5 +64,7 @@ export function splitProfileProperties<T extends Tool>(
 }
 
 function isContentProperty(key: string): boolean {
```
sidvishnoi, line 66: This function and its usage need to be updated to reflect new data structure. Will do separately.

frontend/app/routes/paywall-preview.tsx
```diff
@@ -0,0 +1,104 @@
...
+      } else if (ev.data.action === 'RESET') {
+        el.removeAttribute('hidden')
+        actions.setView('home')
+      }
+    }
+
+    el.updateUI(profile)
+    window.addEventListener('message', listener)
+    return () => {
+      window.removeEventListener('message', listener)
```
kjmitchelljr, line 74: q: would `el.remove()` be called here as well?

sidvishnoi: No. Once the parent element get removed, the child gets removed automatically.

frontend/app/routes/paywall-preview.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const container = document.getElementById('paywall-preview-container')!
+    container.appendChild(el)
+    window.parent.postMessage('ready', window.location.origin)
+
+    const listener = (ev: MessageEvent<Message>) => {
+      if (ev.origin !== window.location.origin) return
+      if (ev.data.action === 'UPDATE') {
+        const profile = ev.data.profile
+        setProfile(profile)
+        document.querySelector(NAME)?.updateUI(profile)
```
DarianM:
```suggestion
        el.updateUI(profile)
```
?

sidvishnoi: Good catch. I had this listener outside earlier.

frontend/app/routes/paywall-preview.tsx
```diff
@@ -0,0 +1,104 @@
...
+        const profile = ev.data.profile
+        setProfile(profile)
+        el.updateUI(profile)
+      } else if (ev.data.action === 'RESET') {
+        el.removeAttribute('hidden')
+        actions.setView('home')
+      }
+    }
+
+    el.updateUI(profile)
```
DarianM, line 71: something is off. why call `updateUI` twice? somewhere in here is a stale closure. (i.e. the paywall UI does not update with the latest profile edits). looking

DarianM (approved):

PR #749, opened 2026-06-02
refactor(frontend): centralize suggested titles

kjmitchelljr:
A part of #718

### Change
Move the tool `suggestedTitles` arrays out of each builder's inline config and into a single file `app/lib/presets.ts`. Builders now import `SUGGESTED_TITLES[TOOL_*]`

Sets up a follow-up analytics change (#718)  to distinguish clicked-preset titles from free-typed text


sidvishnoi (approved):
Looks good. Can you export separate constants instead of an object though? Won't need to import TOOL_ and can simply import like BANNER_SUGGESTED_TITLES in banner. For ALL_TITLES, can do a regular array concat (no object needed).

Also, we should move more such data to centralized places, eventually. Right now, shared/types exports constants (not just types).

PR #751, opened 2026-06-03
feat(InputFieldNumeric): add numeric input; use in paywall editor

### LINKED ISSUE #688: Configure paywall settings
As a publisher, I want to **set a price per article**, so that I control the value of each piece of content I gate. 

Acceptance Criteria: 

- The publisher can set how much their visitors pay for a gated article 

- The currency of the price is not configurable. It is the same currency as the publisher's wallet. 
-----------------------------------------------------------------------------------------------------------------
As a publisher, I want to configure **how much of the screen is visible** and **for how long** before the paywall appears, so that I can give readers enough context to decide whether the content is worth paying for.

Acceptance Criteria: 

- The publisher can set how much of the screen is covered by the paywall, as a percentage.

- The publisher can set a timer in seconds that delays the paywall from appearing after the page loads.

UI

Image

sidvishnoi:
Closes https://github.com/interledger/publisher-tools/issues/688

Add a `InputFieldNumeric` component based on `InputField`, and use it in paywall editor's price and delay inputs.

### DIFF
--- a/frontend/app/components/redesign/components/InputFieldNumeric.tsx
+++ b/frontend/app/components/redesign/components/InputFieldNumeric.tsx
@@ -1,0 +1,84 @@
+import { type ComponentProps, useRef } from 'react'
+import { InputField } from './InputField'
+
+type InputFieldProps = ComponentProps<typeof InputField>
+interface Props extends Omit<InputFieldProps, 'onChange'> {
+  onChange: (value: number) => void
+  min: number
+  max?: number
+  /** Max floating point digits allowed. Set to 0 to allow only integers. */
+  precision?: number
+  value?: string | number
+}
+
+export const InputFieldNumeric = ({
+  value,
+  onChange,
+  min,
+  max,
+  precision = 0,
+  ...props
+}: Props) => {
+  const ref = useRef<HTMLInputElement>(null)
+  const setValue = (value: string) => {
+    ref.current!.value = value
+  }
+
+  const handleChange: InputFieldProps['onChange'] = (ev) => {
+    const val = ev.currentTarget.value.trim()
+    if (!val) return
+    let value = Number(val)
+    if (Number.isFinite(value)) {
+      if (typeof min === 'number' && value < min) value = min
+      onChange(Number(value.toFixed(precision)))
+    }
+  }
+
+  const handleBlur: InputFieldProps['onBlur'] = (ev) => {
+    const value = (() => {
+      const val = Number(ev.currentTarget.value)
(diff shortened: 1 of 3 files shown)


frontend/app/components/redesign/components/InputFieldNumeric.tsx
```diff
@@ -0,0 +1,84 @@
...
+    )
+    if (typeof min === 'number' && newValue < min) return
+    if (typeof max === 'number' && newValue > max) return
+    setValue(newValue.toFixed(precision))
+    onChange(newValue)
+  }
+
+  return (
+    <InputField
+      defaultValue={value}
```
kjmitchelljr, line 73: Right now this value isn't updating/in sync if there is an external change:

1. set price to any value
2. reset changes from the 'Appearance' tab
3. Display/preview shows $2.00 the reset default, but the price box still has the value

frontend/app/components/redesign/components/InputFieldNumeric.tsx
```diff
@@ -0,0 +1,84 @@
...
+  onChange,
+  min,
+  max,
+  precision = 0,
+  ...props
+}: Props) => {
+  const ref = useRef<HTMLInputElement>(null)
+  const setValue = (value: string) => {
+    ref.current!.value = value
+  }
```
DarianM, line 25: afaik using this `ref` is a code smell from our experience with `valtio`, that's making the numeric input uncontrolled and not updating properly. we should use the value from `snap` and change the `profile` only, without using refs as much as possible

frontend/app/components/redesign/components/InputFieldNumeric.tsx
sidvishnoi: So, two things at play here.
1. Resetting Appearance shouldn't have affected price at all.
2. When data is reset, price should've synced with it.

Will address (2) here in this PR, and for (1), see [https://github.com/interledger/publisher-tools/pull/746#discussion_r3341954064](https://github.com/interledger/publisher-tools/pull/746#discussion_r3341954064).

frontend/app/components/redesign/components/InputFieldNumeric.tsx
sidvishnoi: This is unrelated to valtio. It's an input field, managing own state.
I'd have avoided using `ref` (and pass `value={value}`), but then it doesn't sync right due to underlying `InputField` component.

DarianM (approved):

PR #756, opened 2026-06-09
refactor(frontend): edit icon for profile changes in builder accordion

### LINKED ISSUE #287: [Enhancement] Banner/Widget replace check marks with edited icon

kjmitchelljr:
Closes #287

### Changes
Replaces the green check on collapsed builder sections (Banner/Widget/Paywall) with an edit icon that appears only when the section has unsaved changes from the saved baseline

### DIFF
--- a/frontend/app/components/redesign/components/BuilderAccordion.tsx
+++ b/frontend/app/components/redesign/components/BuilderAccordion.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { cx } from 'class-variance-authority'
-import { SVGArrowCollapse, SVGGreenVector } from '@/assets'
+import { SVGArrowCollapse, SVGEdit } from '@/assets'
 import { ToolsSecondaryButton, Divider } from '@/components'
 import { Heading5 } from '@/typography'
 import { GhostButton } from './GhostButton'
@@ -9,7 +9,7 @@ interface Props {
   title: string
   onRefresh: () => void
   onDone?: () => void
-  isComplete?: boolean
+  hasChanges?: boolean
   isOpen?: boolean
   onClick?: (isOpen: boolean) => void
   onToggle?: (e: React.SyntheticEvent<HTMLDetailsElement>) => void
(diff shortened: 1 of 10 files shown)


frontend/app/utils/utils.store.ts
```diff
@@ -96,6 +99,18 @@ export function createToolStoreUtils<T extends Tool>(
...
 
+  function useSectionHasChanges(section: BuilderSection): boolean {
+    const { activeTab } = useSnapshot(toolState)
+    const updates = useSnapshot(store.profilesUpdate)
+    const profile = useSnapshot(store.profiles[activeTab]) as ToolProfile<T>
+    if (!updates.has(activeTab)) return false
+    const baseline = snapshots.get(activeTab)
+    if (!baseline) return false
+    const cur = splitProfileProperties(profile)
+    const base = splitProfileProperties(baseline)
```
sidvishnoi, line 110: We should cache this, as baseline doesn't change often. Splitting on each change would be too much performance wise.

frontend/app/components/redesign/components/BuilderAccordion.tsx
```diff
@@ -50,7 +50,12 @@ export const BuilderAccordion: React.FC<Props> = ({
         )}
       >
         <div className="flex gap-xs items-center">
-          {isComplete && !isOpen && <SVGGreenVector className="w-6 h-[18px]" />}
+          {hasChanges && !isOpen && (
+            <>
+              <span className="sr-only">Edited</span>
+              <SVGEdit className="w-6 h-6 text-text-secondary" />
+            </>
+          )}
```
sidvishnoi, line 58: Let's discuss this in today's team call: whether we want full icon shown, or just the blue dot like we show on profile tabs - as the purpose is same.

sidvishnoi (commented):
I'm happy moving things out of `uiState`, so this is generally in good direction.

kjmitchelljr:
Change was made here: https://github.com/interledger/publisher-tools/pull/770

PR #759, opened 2026-06-09
refactor(frontend): use shared layout, utils on interactive tools

sidvishnoi:
Add `ToolLayoutWithPreview`, which abstracts away most of the layout and setup we use individually in each tool.

Update banner, offerwall, widget and paywall routes to use this layout component.

We pass a lot of things as props to keep this abstraction. Can make it even more abstract, but trying to keep the PR as small as possible for now.
Next, I'd like to get rid of `MobileStepsIndicator`, make `StepsIndicator` adapt to mobile layouts. Then we'd simply have 2-3 steps per tool (the wallet address being first common step), and not require `preview` as a prop. Keep it similar to children esentially.

### DIFF
--- a/frontend/app/components/ToolLayoutWithPreview.tsx
+++ b/frontend/app/components/ToolLayoutWithPreview.tsx
@@ -208,7 +208,7 @@ function BuilderActions({ handleSave }: BuilderActionsProps) {
 
   const onClickScript = async () => {
     setIsLoadingScript(true)
-    await handleSave('script').finally(() => setIsLoading(false))
+    await handleSave('script').finally(() => setIsLoadingScript(false))
   }
 
   return (
--- a/frontend/app/components/ToolLayoutWithPreview.tsx
+++ b/frontend/app/components/ToolLayoutWithPreview.tsx
@@ -198,17 +198,17 @@ type BuilderActionsProps = {
 }
 
 function BuilderActions({ handleSave }: BuilderActionsProps) {
-  const [isLoading, setIsLoading] = useState(false)
-  const [isLoadingScript, setIsLoadingScript] = useState(false)
+  const [isSaving, setIsSaving] = useState(false)
+  const [isGenerating, setIsGenerating] = useState(false)
 
   const onClickSave = async () => {
-    setIsLoading(true)
-    await handleSave('save-success').finally(() => setIsLoading(false))
+    setIsSaving(true)
+    await handleSave('save-success').finally(() => setIsSaving(false))
   }
 
   const onClickScript = async () => {
-    setIsLoadingScript(true)
-    await handleSave('script').finally(() => setIsLoadingScript(false))
+    setIsGenerating(true)
+    await handleSave('script').finally(() => setIsGenerating(false))
   }
 
   return (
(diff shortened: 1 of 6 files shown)


frontend/app/routes/banner.tsx
```diff
@@ -89,218 +72,98 @@ export async function loader({ request, context }: LoaderFunctionArgs) {
   )
 }
 
-export default function Banner() {
+export default function Paywall() {
```
kjmitchelljr:
```suggestion
export default function Banner() {
```

frontend/app/components/ToolLayoutWithPreview.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const [isLoadingScript, setIsLoadingScript] = useState(false)
+
+  const onClickSave = async () => {
+    setIsLoading(true)
+    await handleSave('save-success').finally(() => setIsLoading(false))
+  }
+
+  const onClickScript = async () => {
+    setIsLoadingScript(true)
+    await handleSave('script').finally(() => setIsLoading(false))
```
kjmitchelljr:
```suggestion
await handleSave('script').finally(() => setIsLoadingScript(false))
```

frontend/app/routes/banner.tsx
```diff
@@ -89,218 +72,98 @@ export async function loader({ request, context }: LoaderFunctionArgs) {
...
-                  </div>
-                </div>
-              </div>
-            </div>
-          </div>
-        </div>
-      </div>
+    <div
+      id="preview"
+      className="w-full mx-auto xl:mx-0 xl:sticky xl:top-md xl:self-start xl:flex-shrink-0 xl:w-[504px] h-fit"
```
kjmitchelljr: `ToolLayoutWithPreview.tsx:180-183` wraps the `preview` in `<div id="preview" className="w-full mx-auto ...">` 

Right now it's being double wrapped. Same in `offerwall.tsx`, `paywall.tsx`, and `widget.tsx`. Can most likely remove in those areas I believe to avoid HTML/CSS issues

frontend/app/components/ToolLayoutWithPreview.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+export function ToolLayoutWithPreview({
+  title,
+  description,
+  additionalDescription,
+  walletStore,
+  toolStoreUtils,
+  walletAddressToolName,
+  steps,
+  stepMiddle: stepMiddle,
```
kjmitchelljr: nit: could just pass in `stepMiddle`

frontend/app/components/ToolLayoutWithPreview.tsx
```diff
@@ -0,0 +1,251 @@
...
+    toolStoreUtils.hydrateSnapshotsFromStorage()
+
+    loadState(loaderData.OP_WALLET_ADDRESS)
+    persistState()
+    walletStore.load()
+    walletStore.persist()
+
+    return () => {
+      unsubscribeStorage()
+      unsubscribeUpdates()
```
kjmitchelljr, line 88: AI flagged doing a potential cleanup here for persistState() and walletStore.persist() as multiple listeners are being stacked upon each other. Is something worth further investigating and fixing here, or potential follow-up PR if it does turn out this is something we can improve on?

sidvishnoi: I think we can do as follow up, given with this PR it'll all be in single place (this PR focuses on reducing duplication, not any behaviour changes).

frontend/app/routes/widget.tsx
```diff
@@ -87,199 +66,88 @@ export async function loader({ request, context }: LoaderFunctionArgs) {
...
-                          </div>
-                        </ToolsPrimaryButton>
-                      </div>
-                    </div>
-                  </div>
+    <ToolLayoutWithPreview
+      title="Widget"
+      description={
+        <>
+          The payment widget allows visitors to make one-time payments to
```
DarianM, line 83: nit: do you want to move these into locales? similar to description in paywall.tsx? 
description from banner.tsx also

DarianM (approved):

frontend/app/routes/widget.tsx
sidvishnoi: All in good time...

PR #760, opened 2026-06-09
refactor(frontend): stricter types for session data

sidvishnoi:
Trying to reach `noImplicitAny` state in codebase.

### DIFF
--- a/frontend/app/utils/session.server.ts
+++ b/frontend/app/utils/session.server.ts
@@ -1,7 +1,17 @@
 import { createCookieSessionStorage } from 'react-router'
+import type { PendingGrant, WalletAddress } from '@interledger/open-payments'
+
+type SessionData = {
+  'is-grant-accepted': boolean
+  'is-grant-response': boolean
+  'grant-response': string
+  'payment-grant': PendingGrant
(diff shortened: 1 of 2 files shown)


frontend/app/utils/session.server.ts
```diff
@@ -1,7 +1,17 @@
 import { createCookieSessionStorage } from 'react-router'
+import type { PendingGrant, WalletAddress } from '@interledger/open-payments'
+
+type SessionData = {
+  'is-grant-accepted': boolean
+  'is-grant-response': boolean
+  'grant-response': string
+  'payment-grant': PendingGrant
+  'wallet-address': WalletAddress
+  'validForWallet': WalletAddress['id']
```
kjmitchelljr, line 10: nit: should `validForWallet` be kebab-case or keep as is?

kjmitchelljr (approved):
LGTM 👍🏿

frontend/app/utils/session.server.ts
sidvishnoi: For consistency, yes, should be kebab case. 
But don't want to break any existing sessions. Just adding stricter types instead of changing any behaviour.

PR #763, opened 2026-06-12
fix(PaywallPreview): show in realistic desktop/mobile device size

sidvishnoi:
Finishing touches for https://github.com/interledger/publisher-tools/issues/670


frontend/app/components/paywall/PaywallPreview.tsx
sidvishnoi, line 1: In a follow-up PR, I'll extract the common functionality so it can be used by other tools as well..

sidvishnoi: Done in https://github.com/interledger/publisher-tools/pull/764. But this PR will have to be merged first.

DarianM (approved):
the iframe is a bit small and hard to read on the desktop side

PR #764, opened 2026-06-12
refactor(frontend): add generic iframe-based ToolPreview

sidvishnoi:
So we can reuse this in other tools.


DarianM (approved):

frontend/app/components/ToolPreview.tsx
```diff
@@ -0,0 +1,115 @@
...
+    resizePreview()
+    window.addEventListener('resize', resizePreview)
+    return () => {
+      window.removeEventListener('resize', resizePreview)
+    }
+  }, [iframeSize.width])
+
+  useEffect(() => {
+    const messageHandler = (ev: MessageEvent<MessageFromIframe>) => {
+      if (ev.origin !== location.origin) return
```
adrianboros, line 78: Validate event.source, not only event.origin, because all tool previews share the application origin, a different iframe, popup, or same-origin script can send messages such as CURRENT_SCREEN or READY and have them treated as if they came from this preview. Once this component is reused by multiple tools, every mounted ToolPreview will also receive every same-origin tool message. This can cause cross-tool state changes or handlers processing an incompatible message shape.

Restrict messages to the iframe owned by this component by something like:

```
if (
  ev.origin !== location.origin ||
  ev.source !== iframeRef.current?.contentWindow
) {
  return
}

onMessage(ev.data)
```

frontend/app/components/ToolPreview.tsx
```diff
@@ -0,0 +1,115 @@
...
+      onMessage(ev.data)
+    }
+    window.addEventListener('message', messageHandler)
+    return () => {
+      window.removeEventListener('message', messageHandler)
+    }
+  }, [])
```
adrianboros, line 85: }, [onMessage])

frontend/app/components/ToolPreview.tsx
DarianM: very good point for validating the source !
is this applicable to us tho since we only have one tool render per page all the time?

sidvishnoi: Good points, both.
The thing is, while anyone can send a message to preview, there's only one ToolPreview per route/page. And to send a message to iframe, we've to get a ref to iframe and then postMessage to it.
I added two previews on same page, and only one received message until I pass two different refs to each component.
For cross-tool, each route has its own corresponding preview route, which it sends message to/from. I added paywall preview to banner, and while parent received the message, it had no listener for that message - as that'll require us adding two previews on same page.
So, overall, the current implementation is safe.

PR #773, opened 2026-06-18
fix(/banner): use iframe based preview for banner

sidvishnoi:
Based on https://github.com/interledger/publisher-tools/pull/764, https://github.com/interledger/publisher-tools/pull/763.
We're turning all tools to use iframe for preview.

PR #775, opened 2026-06-18
fix(/offerwall): use iframe based preview for offerwall

sidvishnoi:
Similar to https://github.com/interledger/publisher-tools/pull/773


frontend/app/routes/offerwall.tsx
DarianM: cleanup

DarianM (approved):

PR #776, opened 2026-06-18
fix(widget): use iframe preview, remove preview code from component

sidvishnoi:
Similar to https://github.com/interledger/publisher-tools/pull/773

Additional, removed some preview-mode specific behavior from component, add less custom behavior to CDN version.

### DIFF
--- a/components/src/widget/widget.ts
+++ b/components/src/widget/widget.ts
@@ -161,15 +161,11 @@ export class PaymentWidget extends LitElement {
     if (!this.config) {
       return html``
     }
-    const isPreview = !!this.#controller.isPreviewMode
+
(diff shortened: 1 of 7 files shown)


components/src/widget/widget.ts
DarianM, line 1: contrast-color no longer works for the primary button text

sidvishnoi: Doesn't work in staging or production either 😬

sidvishnoi: I don't see any color contrast stuff in widget - we only added to newer components - offerwall, paywall.

DarianM (approved):

PR #777, opened 2026-06-19
chore: use TypeScript projects/references

sidvishnoi:
Instead of adding TypeScript to each sub-project, use TypeScript's composite and references feature, and use a base tsconfig file for all projects.


package.json
```diff
@@ -15,7 +15,7 @@
     "format:check": "prettier . --check --cache --cache-location='node_modules/.cache/prettiercache'",
     "lint": "eslint . --fix --cache --cache-location 'node_modules/.cache/eslintcache'",
     "lint:check": "eslint . --cache --cache-location 'node_modules/.cache/eslintcache'",
-    "typecheck": "pnpm -r typecheck",
+    "typecheck": "tsc -b",
```
sidvishnoi, line 18: This will be much much faster locally. It doesn't spin up a new `tsc` instance in each project - the single instance knows the rules per project.

frontend/app/entry.server.tsx
```diff
@@ -19,10 +15,6 @@ export default async function handleRequest(
   responseStatusCode: number,
   responseHeaders: Headers,
   reactRouterContext: EntryContext,
-  // This is ignored so we can keep it in the template for visibility.  Feel
-  // free to delete this parameter in your app if you're not using it!
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  loadContext: AppLoadContext,
```
sidvishnoi, line 25: These were caught & ignored by eslint; but our tsconfig also disallowed unused vars (without eslint), so got caught again correctly.

cdn/src/banner.ts
```diff
@@ -41,10 +41,13 @@ function drawBanner(profile: BannerProfile) {
...
   }
 
-  bannerElement.addEventListener('click-extension-link', (e) => {
-    const { link } = (e as CustomEvent<{ link: string }>).detail
-    trackEvent('click_extension_link', { link })
-  })
+  bannerElement.addEventListener(
+    'click-extension-link',
+    (e: CustomEvent<{ link: string }>) => {
+      const { link } = e.detail
```
sidvishnoi, line 47: Earlier, `e` had type `any`. Our strict typescript config didn't apply to this file.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "files": [],
+  "references": [
+    { "path": "./api" },
+    { "path": "./cdn" },
+    { "path": "./frontend" },
+    { "path": "./components" },
+    { "path": "./localenv/s3" },
+    // { "path": "./migrations/s3" },
```
kjmitchelljr, line 9: Out of curiosity, why is s3 path current commented out?

kjmitchelljr (approved):
Looks good overall, great work 👍🏿

tsconfig.json
sidvishnoi: Many changes needed and we'll be getting rid of it soon.

PR #783, opened 2026-06-23
feat(widget): validate wallet provider compatibility

### LINKED ISSUE #498: [FEATURE REQUEST] Widget validation for incompatible wallet provider payment attemps
### Describe the feature you would like to request

Add error handling and user-friendly messages when users attempt payments between incompatible wallet providers in the payment widget  tool.

### Describe the solution you would like

When a user clicks the "Submit" button with an incompatible wallet address:
- Display a clear, actionable error message explaining why the payment cannot proceed
  - like in the wm extension "This page cannot receive payments from your wallet at this time. Please check back later"
- Suggest alternative actions (e.g., use a compatible wallet provider)
- Prevent confusing network errors from reaching the user

## Problem
Currently, when a user attempts to pay from an incompatible wallet provider, the error handling is unclear or non-existent. This can happen in several scenarios:

1. **Testnet to Production**: Attempting to pay from a testnet wallet (e.g., `test.wallet.example`) to a production wallet (e.g., Interledger App wallet)
2. **Unpeered Providers**: Attempting to pay between wallet providers that haven't established peering (e.g., GateHub wallet → Interledger App wallet)

## Implementation Considerations
- potentially new route for wallet provider validation; potentially verification possible by successfully creating a payment quote

kjmitchelljr:
Closes #498

## Summary
Validates wallet compatibility at the payment widget's home view. Incompatible wallets fail immediately on Submit with:

> This page cannot receive payments from your wallet at this time. Please check back later.

## Changes

- **Probe via the existing `POST /payment/quotes`** with `receiveAmount: 1`
- **Error preservation**: `createIncomingPayment` rethrows `OpenPaymentsClientError` as-is so the route handler can classify it
- **Widget integration**: `probeWalletCompatibility` on the `Controller` interface, called in `_handleSubmit` between `getWallet` and the view transition. Has a thin wrapper around `fetchQuote`.


sidvishnoi (changes requested):
Why add a new route instead of expanding error handling in existing quote route?

api/src/routes/payment/validate.ts
```diff
@@ -0,0 +1,53 @@
...
+export type PaymentValidateResult =
+  | { compatible: true }
+  | { error: 'WALLET_MISMATCH' }
+
+// Best-effort metadata pre-check — fail-fast before the probe-quote round-trip.
+// No-op for now; signals tracked in #784.
+const hasIncompatibleMetadata = (
+  _sender: PaymentValidateInput['sender'],
+  _receiver: PaymentValidateInput['receiver'],
+): boolean => false
```
kjmitchelljr: Is this worth exploring - would help with latency, but may be too much to maintain especially if wallet providers do establish peering in the future?

kjmitchelljr:
> Why add a new route instead of expanding error handling in existing quote route?

It was mentioned in the issue: _"potentially new route for wallet provider validation; potentially verification possible by successfully creating a payment quote"_ so I went with that initial approach

kjmitchelljr:
> > Why add a new route instead of expanding error handling in existing quote route?
> 
> It was mentioned in the issue: _"potentially new route for wallet provider validation; potentially verification possible by successfully creating a payment quote"_ so I went with that initial approach

Only other difference/change would be when the validation occurs:
  | When wallet address is submitted (home view) | After amount is entered (initiate view) |
  | :---: | :---: |
  |  |  |

@sidvishnoi

sidvishnoi:
@kjmitchelljr I think it is indeed better to show the incompatible wallet address on home screen (rather than amount screen).
But we don't need a new route in backend for this. We can pass the test amount to existing quote endpoint as well. Just call the quote from home screen that is.

api/src/routes/payment/validate.ts
sidvishnoi: Yes, not worth maintaining.

sidvishnoi:
> > Why add a new route instead of expanding error handling in existing quote route?
> 
> It was mentioned in the issue: _"potentially new route for wallet provider validation; potentially verification possible by successfully creating a payment quote"_ so I went with that initial approach

The issue was likely created before we had the quote endpoint.

api/src/routes/payment/quotes.ts
```diff
@@ -41,6 +42,9 @@ app.post(
           400,
         )
       }
+      if (error instanceof OpenPaymentsClientError) {
+        return json({ error: 'WALLET_UNAVAILABLE' satisfies PaymentError }, 400)
+      }
```
sidvishnoi: `WALLET_UNAVAILABLE` on any other error seems risky - what are other possible cases?
Also, try sending from GateHub (sandbox as well as production) to testnet - is this path triggered or do we go into HTTP 50x?

api/src/utils/open-payments.ts
```diff
@@ -275,6 +274,7 @@ export class OpenPaymentsService {
         },
       )
     } catch (error) {
+      if (error instanceof OpenPaymentsClientError) throw error
       throw createHTTPException(
```
sidvishnoi: Is any other scenario possible? i.e., do we even need the try-catch - if only `OpenPaymentsClientError` is thrown?

cdn/src/widget.ts
```diff
@@ -31,6 +31,16 @@ const drawWidget = (walletAddressUrl: string, profile: WidgetProfile) => {
...
     getWallet: (walletAddressUrl) => getWallet(API_URL, walletAddressUrl),
+    async probeWalletCompatibility({ sender, receiver }) {
+      const result = await fetchQuote(API_URL, {
+        sender,
+        receiver,
+        receiveAmount: 1,
+      })
+      return 'error' in result && result.error === 'WALLET_UNAVAILABLE'
+        ? { ok: false, code: 'WALLET_UNAVAILABLE' }
+        : { ok: true }
```
sidvishnoi: What if we throw with error message instead when not-ok? If it doesn't throw, all is good, if it throws, use error message.

components/src/widget/widget.ts
```diff
@@ -61,9 +61,19 @@ export class PaymentWidget extends LitElement {
...
       const walletInfo = await this.#controller.getWallet(walletAddress)
+      const receiver = await this.#receiver
+
+      const probe = await this.#controller.probeWalletCompatibility({
+        sender: walletInfo,
+        receiver,
+      })
+      if (!probe.ok) {
+        return errorMessageFor(probe.code)
+      }
```
sidvishnoi: With throw, it can become like following:
```ts
try {
  await this.#controller.probeWalletCompatibility({ ... })
  // if it doesn't throw, can send
} catch(err) {
  return err.message
}
```

api/src/routes/payment/quotes.ts
kjmitchelljr: Hey @sidvishnoi, yes that was something I've looked into. I only get the 50x error with different wallets. Right now there's no specific error code/message for unpeered wallets.

api/src/utils/open-payments.ts
kjmitchelljr: @sidvishnoi good catch, I actually removed the try/catch in a previous iteration, but wasn't sure if we wanted to keep the _'Unable to create incoming payment'_ message specifically for this the create method OR if it'd be fine to use the OpenPaymentsClientError handler.

But I agree with the change.

components/src/widget/widget.ts
kjmitchelljr: I'll try this out 👍🏿

api/src/routes/payment/quotes.ts
kjmitchelljr: Tested again, I received this after removing the try/catch which I believe is expected behavior:
```
▲ [WARNING] [payment/quotes] OpenPaymentsClientError shape: {

    status: 400,
    code: '400',
    description: 'invalid receiver',
    message: 'Error making Open Payments POST request'
  }
```

sidvishnoi (approved):

PR #785, opened 2026-06-24
feat(frontend): use `subject` grant for wallet ownership check

kjmitchelljr:
Closes #638

  ## Summary
  - Save-profile flow now requests a GNAP `subject` grant first to confirm wallet ownership (Rafiki v2.1.0+), so the IDP consent screen reads as an ownership prompt
  - If the wallet's auth server rejects the subject grant (older Rafiki without `subject` support), we automatically fall back to the previous outgoing-payment grant
  - `isGrantValidAndAccepted` accepts either a finalized subject grant or a finalized access-token grant.

### DIFF
--- a/frontend/app/utils/open-payments.server.ts
+++ b/frontend/app/utils/open-payments.server.ts
@@ -85,6 +85,15 @@ async function createClient(env: Env) {
   return client
 }
 
+// authServers known to support subject (remove once broadly supported)
+const SUBJECT_GRANT_AUTH_SERVERS = new Set<string>([
+  'auth.interledger-test.dev',
+])
+
+function supportsSubjectGrant(walletAddress: WalletAddress): boolean {
+  return SUBJECT_GRANT_AUTH_SERVERS.has(new URL(walletAddress.authServer).host)
+}
+
 export async function createInteractiveGrant(
   env: Env,
   args: {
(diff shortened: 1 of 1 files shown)


frontend/app/utils/open-payments.server.ts
```diff
@@ -93,22 +94,60 @@ export async function createInteractiveGrant(
...
+      redirectUrl: args.redirectUrl,
+    })
+  }
+}
+
+async function createSubjectGrant(params: {
+  walletAddress: WalletAddress
+  nonce: string
+  opClient: AuthenticatedClient
+  redirectUrl?: string
```
sidvishnoi: Let's make redirectUrl required.

sidvishnoi (approved):
Can you check on existing wallets if we do need this fallback now? There are only 3 wallets - Testnet, Interledger, GateHub, if each supports subject grants now, we can let go of the fallback. But good otherwise.

frontend/app/utils/open-payments.server.ts
kjmitchelljr: Change made here. [bdfd522](https://github.com/interledger/publisher-tools/pull/785/commits/bdfd52200abaf42687f26a7a9d5cf5f490ac91a0)

Confirmed it's working in Testnet. Waiting for verification for Gatehub wallet. Going through setup for Interledger wallet now. Will merge once confirmed behavior.

frontend/app/utils/open-payments.server.ts
```diff
@@ -88,27 +89,65 @@ export async function createInteractiveGrant(
...
+
+  try {
+    return await createSubjectGrant({
+      walletAddress: args.walletAddress,
+      nonce: clientNonce,
+      opClient,
+      redirectUrl: args.redirectUrl,
+    })
+  } catch {
+    return createOutgoingPaymentGrant({
```
DarianM: this fallback method won't work because there is no context left to catch the error if something goes wrong or the IDP doesn't support the `subject` field (i.e. the error occurs in the context of the redirect page).

I think we need to conditionally check for wallet providers in WA, then use "createSubjectGrant" only where we know it is supported, then gradually deprecate it as support is added

kjmitchelljr: Added conditional check in this [36c26ca](https://github.com/interledger/publisher-tools/pull/785/commits/36c26caae8cbbfdcae3f19089e40a672d0e7c62d)

sidvishnoi (approved):

DarianM (approved):

PR #795, opened 2026-06-30
chore(dep): move pnpm override

kjmitchelljr:
Resolves warning: _The "pnpm" field in package.json is no longer read by pnpm. The following keys were ignored: "pnpm.overrides". See https://pnpm.io/settings for the new home of each setting._


sidvishnoi:
Let's do it when we upgrade to pnpm v11.

PR #796, opened 2026-06-30
chore(dep): upgrade to pnpm 11+

kjmitchelljr:
## Summary
  - Bumps the package manager from pnpm 10.34.3 → 11.9.0
  - Migrates `pnpm-workspace.yaml` from the removed `onlyBuiltDependencies` list to the new `allowBuilds` map

Closes #728

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,7 +6,6 @@ settings:
 
 overrides:
   esbuild: ^0.28.1
-  wrangler: ^4.100.0
 
 importers:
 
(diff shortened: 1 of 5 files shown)


pnpm-lock.yaml
```diff
@@ -5,7 +5,6 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
-  esbuild: ^0.28.1
```
sidvishnoi, line 8: Let's enforce override for esbuild as well (I think it came from `resolutions` in package.json earlier). Without it, this lockfile is adding so many esbuild versions.

kjmitchelljr: Will do 👍🏿

sidvishnoi: @kjmitchelljr You don't need to reply to confirm simple fixes. I get quite a few notifications, so just sending commits is perfect. Let's use comments if we need to discuss or clarify anything.

sidvishnoi (approved):

PR #809, opened 2026-07-07
fix(paywall): apply figma specs and coverage driven layout

### LINKED ISSUE #808: Paywall design updates (Phase 4)
## Scope

Update the current Paywall design to match Phase 4 of the Figma design spec. Minor updates include:

* Updating default text
* Adding Inter font as an option for users
* Fixing some padding and margin  paywall screen
* Having the lock icon only for 75% and 100% coverages
* Fixing iOS issue where user was still able to scroll once clicking keyboard
* interledger/publisher-tools#781 work also completed in update

## Delivered in

PR [#809](<https://github.com/interledger/publisher-tools/issues/809>)

kjmitchelljr:
> **Split into three PRs** — merge in this order:
  >
  > 1. **https://github.com/interledger/publisher-tools/pull/816** — `fix(paywall): lock underlying page scroll and use dvh for coverage`
  > 2. **#https://github.com/interledger/publisher-tools/pull/815** — `feat(paywall): add Inter as a selectable font family`
  > 3. **This PR** — rebased onto `main` after 1 & 2 land

## Summary
- Minor updates to paywall design to adhere to Figma spec
- Added link (button) that navigates to the `payStart` view if a user has already paid for the article
- Added styling rules and conditions based on the paywall coverage selected, e.g. whether to show icon, centering of content, padding, etc.

Closes #808


components/src/paywall/index.ts
```diff
@@ -253,6 +253,25 @@ export class Paywall extends LitElement {
...
+    const spacious = coverage.value >= 50
+    this.style.setProperty(
+      '--wmt-content-justify',
+      spacious ? 'center' : 'flex-start',
+    )
+    this.style.setProperty(
+      '--wmt-content-align',
+      spacious ? 'center' : 'flex-start',
+    )
+    this.style.setProperty('--wmt-content-gap', spacious ? '1.5rem' : '0.75rem')
```
sidvishnoi: Use CSS.

components/src/constants.ts
```diff
@@ -8,6 +8,21 @@ export const FONT_MAP = new Map<
...
+        'ui-sans-serif',
+        'system-ui',
+        '-apple-system',
+        'BlinkMacSystemFont',
+        '"Segoe UI"',
+        'Roboto',
+        'sans-serif',
+      ],
+    },
+  ],
```
sidvishnoi, line 25: This is not in product requirements, at least as of yet. We're using Inter only in our editor UI, not in embeds.

components/src/paywall/index.ts
sidvishnoi: Mobile (50% coverage). Note the difference in two screens.

| Home | Form |
| --- | --- |
| Image | Image |

components/src/paywall/index.ts
```diff
@@ -253,6 +253,25 @@ export class Paywall extends LitElement {
     applyFontFamily(this, font.name, 'paywall', fontBaseUrl)
     this.dataset.fontSize = font.size
     this.style.setProperty('--wmt-height', `${coverage.value}vh`)
+    this.style.setProperty(
+      '--wmt-icon-display',
+      coverage.value >= 75 ? 'grid' : 'none',
```
sidvishnoi: Why hide icon on 50% coverage as well?

components/src/paywall/components/home.ts
```diff
@@ -27,6 +27,10 @@ export class PaywallHome extends LitElement {
       <div class="top">
         <h2 class="title">${this.title}</h2>
         <p class="description">${this.description}</p>
+        <p class="description">
+          Already paid? Enter your wallet address to access this content for
+          free.
+        </p>
```
sidvishnoi, line 34: Should this be on the form screen as well?

components/src/paywall/components/common.css
```diff
@@ -18,22 +19,15 @@ p {
   align-self: center;
   width: 100%;
   min-height: 0;
-  --min-height-threshold: 6rem;
-  height: clamp(
-    0%,
-    (100% - var(--min-height-threshold)) * 999,
-    var(--coverage) * 0.3
-  );
```
sidvishnoi, line 26: This ensured the icon size scaled to screen size, font size, virtual keyboard (phone) and available space, without hardcoding different settings.

sidvishnoi (changes requested):

components/src/paywall/components/common.css
kjmitchelljr: Which direction would you want to go? Made the switch to use flex because there was an empty gap between description and price due to the initial grid layout and the hidden icon leaving the row behind. I've tested and the size still looks to scale with `max-height: calc(var(--coverage) * 0.3);` we just don't have the auto collapse, but in this PR it has been replaced with the coverage-driven hiding of the image at 25% and 50%

**Staging:**
Screenshot 2026-07-09 at 1 19 07 PM

sidvishnoi: As long as matches Figma, I'm ok with either.

components/src/paywall/components/common.css
```diff
@@ -49,9 +43,11 @@ button {
 
   background-color: var(--theme);
   /* contrast-color() is unreliable in Chrome shadow DOM; approximate with OKLCH:
-     https://css-tricks.com/approximating-contrast-color-with-other-css-features/ */
+     https://css-tricks.com/approximating-contrast-color-with-other-css-features/
+     Threshold shifted to L=0.9 so common brand colors (e.g. teal L≈0.73) get
+     white text; only near-white themes fall back to black. */
   color: white;
-  color: oklch(from var(--theme) round(1.21 - L) 0 0);
+  color: oklch(from var(--theme) round(1.4 - L) 0 0);
```
sidvishnoi: I won't recommend this. It might be more brand-matching, but can cause a11y issues with others.
I saw brand-teal bg with white text has only 2.48 contrast ratio. Even if "black" text might look odd, it's visually and mathematically more legible.

components/src/paywall/components/form.ts
```diff
@@ -69,17 +72,18 @@ export class PaywallWalletAddressForm extends LitElement {
...
 
-        <button type="submit" ?disabled=${this._loading}>
-          ${
-            this._loading
-              ? html`<wm-dots-loader></wm-dots-loader
-                  ><span class="sr-only">${this.ctaText}</span>&nbsp;`
-              : this.ctaText
-          }
-        </button>
+          <!-- Note: the form step always says "Unlock", independent of the configurable CTA text on the home view. -->
```
sidvishnoi: This comment is not needed - it's clear from text.
I'd recommend passing `ctaText="Unlock"` from parent for this, and keep the diff small.

shared/types/index.ts
```diff
@@ -268,6 +268,7 @@ export const WIDGET_POSITION = {
 export type WidgetPositionKey = keyof typeof WIDGET_POSITION
 
 export const FONT_FAMILY_OPTIONS = [
+  'Inter',
```
sidvishnoi, line 271: Inter is the last font as well. Duplicated font.

sidvishnoi (changes requested):

sidvishnoi (approved):
LGTM. Will recommend @DarianM, @MichaelKohler to review as well before merging.

MichaelKohler (approved):

Issue #480, opened 2025-12-11
[improvement] better handle state flow in grant interaction page


DarianM:
The `payment-confirmation.tsx` route currently only handles successful grant interactions. It doesn't properly handle:

1. **Declined interactions** - when user declines the grant consent
2. **Error states** - when something goes wrong during the interaction
3. **Misleading naming** - file is named "payment-confirmation" but actually handles "grant interaction" (Open Payments flow)

Currently, all scenarios show the success UI with checkmark, even if the interaction failed or was declined by the user.

Goals
- [ ]  rename route to better reflect its purpose (e.g. `grant-interaction`)
- [ ] handle declined interaction state (when user clicks "Deny" in consent screen)
- [ ] handle error states (network errors, invalid responses, etc)
- [ ] display appropriate UI for each state
- [ ] pass the correct state back to whomever uses

Note:

- similar flows and logic like welcome pages the extension uses `webmonetization.org/welcome/?result=[...]&intent=[...]`

PR #811, opened 2026-07-10
fix(frontend/grant): handle more states in wallet ownership

kjmitchelljr:
## Follow-up (#480 keep open)
Issue #480 targets `frontend/app/routes/payment-confirmation.tsx` — the popup shown to a *user* after they complete the wallet consent step during a widget payment. That page still shows the success UI regardless of decline/error result

This PR instead fixes a bug in the *publisher-side* wallet connect flow, where non-happy paths surfaced an empty error message that fell through to `StatusDialog`'s default of *"Your edits have been saved"*.

  ## Changes
  Every outcome sets an explicit, user-facing message:
  - Declined → *"Wallet authorization was declined."*
  - Missing session / `interact_ref` → *"Your session has expired. Please try again."*
  - Continuation throws → *"Unable to verify wallet access. Please try again."*
  - Continuation returns non-finalized → *"Wallet authorization could not be completed. Please try again."*
  - Accepted → existing *"Wallet ownership confirmed!"*
  
  ## Tested

  Verified with `interledger-test.dev` wallet on `http://localhost:3000/tools/banner`:
  - [x] Success → *"Wallet ownership confirmed!"*
  - [x] Deny/Cancel → *"Wallet authorization was declined."*
  - [x] Incognito → *"Your session has expired. Please try again."*
  - [x] Bad `interact_ref` → *"Unable to verify wallet access. Please try again."*


frontend/app/routes/api.grant.$type.ts
```diff
@@ -15,25 +15,28 @@ export async function loader({ params, request, context }: Route.LoaderArgs) {
...
-  let grantResponse = result == 'grant_rejected' ? 'Grant was declined' : ''
   let isGrantAccepted = false
+  let grantResponse: string
 
-  if (walletAddress && grant && interactRef) {
+  if (result === 'grant_rejected') {
+    grantResponse = 'The grant request was declined.'
+  } else if (!walletAddress || !grant || !interactRef) {
+    grantResponse = 'Grant session is missing or expired. Please try again.'
+  } else {
```
sidvishnoi, line 26: Please move these messages to the locales file.
Also, avoid using the word "grant" if possible.

sidvishnoi (commented):
While this doesn't address https://github.com/interledger/publisher-tools/issues/480, this is a welcome change. Please update PR description to reflect the issue you're resolving.

frontend/app/routes/api.grant.$type.ts
```diff
@@ -1,8 +1,11 @@
 import { redirect } from 'react-router'
+import en from '~/i18n/locales/en.json'
 import { isGrantValidAndAccepted } from '~/utils/open-payments.server'
 import { commitSession, getSession } from '~/utils/session.server'
 import type { Route } from './+types/api.grant.$type'
 
+const messages = en.grantInteraction
+
```
sidvishnoi: I understand we can't use the `useTranslation` hook in `loader`, but we can return the "key" from loader, and `useTranslation` based on result of `useLoaderData`. Then we won't need two types of `useTranslation` handling in same app. This will also save the cookie storage (as `grantResponse`) is part of the session cookie.

sidvishnoi (approved):

PR #816, opened 2026-07-15
fix(paywall): iOS lock page scroll and add dvh for coverage

kjmitchelljr:
## Summary

- Fixes an issue found in testing where the user was still able to scroll beneath the underlying contents of the page that had the paywall. Found when the keyboard appeared upon entering the wallet address to pay for the content.
- Switched to [`dvh`](https://medium.com/@tharunbalaji110/understanding-mobile-viewport-units-a-complete-guide-to-svh-lvh-and-dvh-0c905d96e21a) to adjust properly when user opens keyboard. Also, added fallback for `vh` for older devices
- Added abilitiy for paywall to be scrollable in cases where the content of the paywall exceeds the visible area

## Tested
- Tested on iOS Simulator and on physical iPhone 17 device
- Tested to make sure that desktop and tablet implementations weren't affected by changing `document.body` CSS properties

Split from #809. Related to #808.

### DIFF
--- a/components/src/paywall/index.ts
+++ b/components/src/paywall/index.ts
@@ -285,12 +285,10 @@ export class Paywall extends LitElement {
     } = this.#config
 
     const fontBaseUrl = new URL('/assets/fonts/', this.#controller.cdnUrl).href
-    const vhUnit = CSS.supports('height', '100dvh') ? 'dvh' : 'vh'
 
     applyFontFamily(this, font.name, 'paywall', fontBaseUrl)
     this.dataset.fontSize = font.size
-    this.style.setProperty('--wmt-height', `${coverage.value}${vhUnit}`)
-    this.style.setProperty('--wmt-viewport-height', `100${vhUnit}`)
+    this.style.setProperty('--wmt-height', `${coverage.value}`)
     this.style.setProperty('--wmt-background', colors.background as string)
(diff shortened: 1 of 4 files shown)


components/src/paywall/index.ts
```diff
@@ -250,9 +285,12 @@ export class Paywall extends LitElement {
     } = this.#config
 
     const fontBaseUrl = new URL('/assets/fonts/', this.#controller.cdnUrl).href
+    const vhUnit = CSS.supports('height', '100dvh') ? 'dvh' : 'vh'
```
sidvishnoi: Let's do this with CSS `@supports` rule.

components/src/paywall/index.ts
```diff
@@ -250,9 +285,12 @@ export class Paywall extends LitElement {
...
+
     applyFontFamily(this, font.name, 'paywall', fontBaseUrl)
     this.dataset.fontSize = font.size
-    this.style.setProperty('--wmt-height', `${coverage.value}vh`)
+    this.style.setProperty('--wmt-height', `${coverage.value}${vhUnit}`)
```
sidvishnoi: We can pass `--wmt-height: 50` (instead of 50vh/50dvh) and handle it in `calc` in CSS. Let's do as much of styling via CSS instead of JS where possible.

sidvishnoi (approved):

PR #817, opened 2026-07-15
fix(frontend): update label for corner radius

kjmitchelljr:
## Summary

Minor update to label for corner radius in tool configurations to read "Button Corner Radius"

Found while testing #809


sidvishnoi (changes requested):
I don't think it's desired for all tools. Banner for example has a container radius, no button.

kjmitchelljr:
No need to have for other tools. Only Paywall does the button corner radius change which is already in #809

PR #828, opened 2026-07-17
chore(api): better commands to run d1 database migrations

sidvishnoi:
Instead of running each script manually, use `wrangler d1 migrations apply` command.
I've ran this on production database.


MichaelKohler (approved):
`api/migrations/README.md` should be adjusted as well

PR #832, opened 2026-07-21
chore(dep): upgrade tailwind to v4

kjmitchelljr:
## Summary
- Upgrades the Publisher Tools `frontend` to Tailwind v4 - reference the [Upgrade guide](https://tailwindcss.com/docs/upgrade-guide) for the changes made.
- A lot of small changes were mainly renamed utilities and a few removed deprecated utilities
- Chose to use the Vite plugin vs. PostCSS since the project uses Vite

Closes #800

PR #837, opened 2026-07-21
chore(frontend): Tailwind v4 cleanup

kjmitchelljr:
## Summary
Follow-up PR to cleanup some of the Tailwind v4 migration and implementing consistent design pattern in repo

1. Fixes an issue with `max-w-md` that had previously caused `payment-confirmation.tsx` to break (resolved in #832), but made note for why we're going with arbitrary values
2. Dropped our custom `--radius-*` block and will just use [Tailwind v4 defaults](https://tailwindcss.com/docs/border-radius) - `rounded-lg` now resolves to 8px
3. Fix for inconsistent value vs. token usage for `p/m/gap/space-*` - updated to semantic tokens across the board. There still exists some off-tokens, (e.g. `p-10` in `frontend/app/root.tsx`) because it doesn't match any of the semantic ones we have. Could add new ones in future work.
4. Sizing convention consistency, for `w/h/size-*` using `[px]` - right now it doesn't apply for small icons, e.g. `w-4/5/6` I left those numeric for now
5. Clean up of leftover v3 convention and simplified the hero image in `frontend/app/routes/_index.tsx`

## Test
- [x] Build still compiled without warning
- [x] Visual smoke test on all pages compared with production
- [x] Verified `max-w-*` sites render properly


kjmitchelljr:
Ready for review 👍🏿

sidvishnoi (changes requested):
Noticed some visual regressions, two in spacing; and one icon size.

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |

Left (before), right (after) if this screenshot helps better:
image

MichaelKohler:
@sidvishnoi I did some minor changes, but did not want to spend much time fixing this. I think given the current strategy, this is close enough. Would you agree?

sidvishnoi (approved):

PR #840, opened 2026-07-22
fix(frontend): add failure handling to grant interaction flow

### LINKED ISSUE #480: [improvement] better handle state flow in grant interaction page

kjmitchelljr:
## Summary
- Rename `payment-confirmation` route to `grant-interaction` as that better reflects the purpose page which is handling the result of the grant interaction
- Addition of failure view
- Update to `api/src/routes/payment/redirect.ts` to redirect to page if there is an error

### Tested
- [x] Payment successful - confirmed the success view still shows correctly
- [x] Payment failed - confirmed that the failure view shows

Closes #480


frontend/app/routes/grant-interaction.tsx
```diff
@@ -0,0 +1,47 @@
...
+  Icon: ComponentType<SVGProps<SVGSVGElement> & { className: string }>
+  title: string
+  message: string
+}
+
+const SUCCESS_VIEW: View = {
+  Icon: SVGMarkSuccess,
+  title: 'Payment Complete!',
+  message:
+    'Your payment has been processed successfully.\nYou can now close this window.',
```
sidvishnoi: Let's move the strings to i18n file.

frontend/app/routes/grant-interaction.tsx
```diff
@@ -0,0 +1,47 @@
+import type { ComponentType, SVGProps } from 'react'
+import { useSearchParams, type MetaFunction } from 'react-router'
+import { SVGErrorVector, SVGMarkSuccess } from '@/assets'
+
+export const meta: MetaFunction = () => {
+  return [
+    { title: 'Grant Interaction' },
+    { name: 'description', content: 'Grant interaction result' },
```
sidvishnoi: We can set these based on `result` as well.

api/src/routes/payment/redirect.ts
```diff
@@ -59,46 +71,46 @@ app.get(
           { status: 'GRANT_REJECTED' },
           { expirationTtl: 5 * 60 /* 5 minutes */ },
         )
-        return redirect(
-          urlWithParams(data.redirectUrl, {
-            paymentId,
-            result: 'failure',
-          }),
-        )
+        return redirectFailure(redirect, data.redirectUrl, paymentId)
```
sidvishnoi: I don't see benefit in this abstraction. It's simply `redirect(url)`.
Maybe abstract this like `redirect(getRedrirectUrl(status))` and define `const getRedirectUrl = (result) => urlWithParams(data.redirectUrl, { paymentId, result })` within the handler.

api/src/routes/payment/redirect.ts
```diff
@@ -59,46 +71,46 @@ app.get(
...
-            sender: data.sender,
-            receiver: data.receiver,
-            amount: data.amount,
-            outgoingPaymentGrantAccessToken: accessToken,
-          },
-          { expirationTtl: 5 * 60 /* 5 minutes */ },
-        )
-        return redirect(
-          urlWithParams(data.redirectUrl, {
+        try {
```
sidvishnoi: Nesting `try {} catch {}` makes it harder to follow. Why not handle the error in main catch?

frontend/app/i18n/locales/en.json
```diff
@@ -30,7 +30,17 @@
...
+    "payment.meta.successTitle": "Payment Complete",
+    "payment.meta.successDescription": "Your payment has been processed successfully.",
+    "payment.meta.failureTitle": "Payment not authorized",
+    "payment.meta.failureDescription": "The payment was declined or could not be completed.",
+    "payment.meta.defaultTitle": "Grant Interaction",
+    "payment.meta.defaultDescription": "Grant interaction result",
+    "payment.success.title": "Payment Complete!",
+    "payment.success.message": "Your payment has been processed successfully.\nYou can now close this window.",
+    "payment.failure.title": "Payment not authorized",
+    "payment.failure.message": "The payment was declined or could not be completed.\nYou can now close this window."
```
sidvishnoi: These messages belong to only the `grant-interaction` page (when user makes payment using widget), and are unrelated to grantInteraction when we save config.

frontend/app/routes/grant-interaction.tsx
```diff
@@ -0,0 +1,64 @@
+import type { ComponentType, SVGProps } from 'react'
+import { useSearchParams, type MetaFunction } from 'react-router'
+import { SVGErrorVector, SVGMarkSuccess } from '@/assets'
+import en from '~/i18n/locales/en.json'
+import { useTranslation } from '~/i18n/useTranslation'
+
+export const meta: MetaFunction = ({ location }) => {
+  const params = new URLSearchParams(location.search)
+  const result = params.get('result')
+  const gi = en.grantInteraction
```
sidvishnoi: Nit: so we can know this is part of "translations" - `gi` is very random.
```suggestion
  const t = en.grantInteraction
```

api/src/routes/payment/redirect.ts
```diff
@@ -35,12 +35,14 @@ app.get(
...
+    let data: Awaited<ReturnType<typeof getData>> | undefined
+    const { paymentId } = req.valid('param')
+
     try {
       const openPayments = await OpenPaymentsService.getInstance(env)
-      const { paymentId } = req.valid('param')
       const queryParams = req.valid('query')
 
-      const data = await getData(env.PUBLISHER_TOOLS_KV, paymentId)
+      data = await getData(env.PUBLISHER_TOOLS_KV, paymentId)
```
sidvishnoi: We can do the `data = await getData(env.PUBLISHER_TOOLS_KV, paymentId)` outside of try-catch as well. It won't throw (if it throws, it'll end up on same HTTP 50x path, just not as a custom error).

sidvishnoi (changes requested):

sidvishnoi (approved):

PR #860, opened 2026-08-17
chore(deps): update dependencies

MichaelKohler:
This updates all the dependencies except for `typescript`, which we can't update due to `typescript-eslint`. Includes an upgrade to eslint 10.


sidvishnoi (approved):
Don't forget to run `pnpm dedupe` if not run already.

PR #861, opened 2026-08-17
chore(deps): update eslint in frontend

MichaelKohler:
I had forgotten about the frontend eslint 10 upgrade in the previous PR.

Additionally we now can also set `tailwindcss` to 4.3.3.


sidvishnoi (approved):
As a follow up, let's move eslint, esbuild, wrangler, typescript to workspace dependencies, instead of keeping a version in each package.

